PR #19, opened 2015-01-03
Add user profiles to database model

gunthercox:
Until this point chatterbot has used flat files to store statements from conversations. This pull request implements database storage which will allow several useful changes to be implemented in order to improve results.

- [x] The database setup allow all possible responses for a known statement to be queried.
- [x] The number of times for which statements occur is now being recorded. This is used to weight response results such that more common responses are chosen.
- [ ] Profiles have been created and are are used to track statements said by each user. This will be an element that can be used in the future to gauge potential sarcasm and also to return more relevant responses by making them user-specific.
- [ ] Testing has been rebuilt to account for these changes

The living document detailing some of the changes that will be happeing here can be edited at  https://docs.google.com/document/d/1AWPgrpEyzs2T1gm3Vlp-oBaoMpgO860GfDFWcqCq5To/edit#

### DIFF
--- a/chatterbot/algorithms/engram.py
+++ b/chatterbot/algorithms/engram.py
@@ -1,43 +1,37 @@
-def get_closest(text, log_directory, similarity_threshold):
-    from chatterbot.conversation import Statement, Conversation
-    import os
-
-    closest_response = []
-    closest_ratio = 0
-
-    for log in os.listdir(log_directory):
-        path = log_directory + "/" + log
-
-        if os.path.isfile(path):
-            conversation = Conversation()
-            conversation.read(path)
-            response, ratio = conversation.find_closest_response(text)
-
-            # A response may not be returned if a log has less than one line
-            if response:
-                if ratio > closest_ratio:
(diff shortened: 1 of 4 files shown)


chatterbot/algorithms/engram.py
```diff
@@ -1,43 +1,38 @@
...
+                if statement["occurrences"] > closest_statement["occurrences"]:
+                    closest_statement = statement
 
-    if not closest_response:
-        from chatterbot.conversation import Statement
-        default = Statement("Error", "No possible replies could be determined.")
-        return [default]
+    # Select all statements which have been used to respond to this object
+    possible_responses = statements.find(
+        {"in_response_to": ObjectId(closest_statement["_id"])}
```
kevin-brown: Is there a reason why you are casting this to an `ObjectId`? I'm pretty sure it's not needed, and you aren't casting it in other places.

gunthercox: Nope, this isn't needed.

PR #23, opened 2015-01-05
Newmaster

PaarthBatra:
added some small changes to original code to fix some bugs . Although those functions were not being used . Still helps in keeping the code clean .

### DIFF
--- a/chatterbot/conversation.py
+++ b/chatterbot/conversation.py
@@ -58,7 +58,7 @@ class Statement(object):
         """
         self.response_to = previous_statement
 
-    def detect_sentiment():
+    def detect_sentiment(self):
         """
         A property that describes hows the 
         """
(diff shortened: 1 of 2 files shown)


chatterbot/conversation.py
```diff
@@ -216,7 +216,7 @@ def get_sentiment(name):
         conversation.
         """
         sentiment = []
-        for statement in self:
+        for statement in name:
```
kevin-brown: This was set to `self` before because the `Conversation` object [defines an `__iter__` method](https://github.com/PaarthBatra/ChatterBot/blob/2de82a504ecdeae2790ba7226a2f86dbfff971dc/chatterbot/conversation.py#L90-L95) that yields each statement. This basically allowed you to get each statement in the conversation in a loop.

`name` is used to only get statements for a specific person in the conversation, and iterating over it will give you each individual character.

gunthercox:
Did you have any more changes to make in this pull?

PaarthBatra:
Not yet .

PR #39, opened 2015-05-03
various fixes: clean whitespace, fix typos

justindomingue:


chatterbot/__init__.py
```diff
@@ -82,12 +82,12 @@ def train(self, conversation):
 
             statement = conversation[i]
 
+            database_values = self.database.find(statement)
```
justindomingue: Only call `database.find()` once.

chatterbot/__init__.py
```diff
@@ -82,12 +82,12 @@ def train(self, conversation):
...
+
             # Create an entry if the statement does not exist in the database
-            if not self.database.find(statement):
+            if not database_values:
                 self.database.insert(statement, {})
 
-            database_values = self.database.find(statement)
```
kevin-brown: I _think_ this was there so the search could be re-evaluated. Otherwise `database_values` is still `None` (like it was before).

justindomingue: Yes, I completely agree

gunthercox:
Hi, sorry for the late reply. The changes look good and I've merged them onto master.
I'm going to go ahead with @kevin-brown's recommendation and have the database adapter return the newly inserted statement.

PR #44, opened 2015-06-06
Add closest-match logic adapter

gunthercox:
I've started looking into using [NLTK](http://www.nltk.org/) to compare two statements by the meaning of their content.
**Closest Match**
- [x] Determine close matches based on the nearness of the meaning of word synsets.

### DIFF
--- a/chatterbot/adapters/logic/closest_meaning.py
+++ b/chatterbot/adapters/logic/closest_meaning.py
@@ -1,0 +1,105 @@
+from .logic import LogicAdapter
+
+from nltk.corpus import wordnet
+from nltk.corpus import stopwords
+from nltk import word_tokenize
+
+
+class ClosestMeaningAdapter(LogicAdapter):
+
+    def __init__(self):
+        super(ClosestMeaningAdapter, self).__init__()
+        from nltk.data import find
+        from nltk import download
+
+        # Download data if needed
+
+        try:
+            find('wordnet.zip')
+        except LookupError:
+            download('wordnet')
+
+        try:
+            find('stopwords.zip')
+        except LookupError:
+            download('stopwords')
+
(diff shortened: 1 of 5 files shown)


chatterbot/adapters/logic/closest_meaning.py
```diff
@@ -0,0 +1,94 @@
+from .logic import LogicAdapter
+from collections import Counter
+import nltk
```
kevin-brown: In your tests you should be installing `nltk`, and you probably want to guard against the `ImportError` here and raise a useful message.

No need to require it in your `setup.py`, as it's only needed for this adapter.

chatterbot/adapters/logic/closest_meaning.py
```diff
@@ -0,0 +1,94 @@
...
+        is True.
+        """
+        lower = text.lower()
+        tokens = word_tokenize(lower)
+
+        # Remove any stop words from the string
+        if exclude_stop_words:
+            excluded_words = stopwords.words("english")
+
+            for token in tokens:
```
kevin-brown: You might want to consider using sets here.

`set(tokens) - set(excluded_words)`

chatterbot/adapters/logic/closest_meaning.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_similarity(self, string1, string2):
+        """
+        Calculate the similarity of two statements.
+        """
+
+        tokens1 = self.get_tokens(string1)
+        tokens2 = self.get_tokens(string2)
+
+        print tokens1, tokens2
```
kevin-brown: `print(token1, tokens2)` if you want tests to run on Py3.

PR #53, opened 2015-09-10
no need to multiply by 'count'

zan-kusterle:
values = list(statement)[0]
IndexError: list index out of range


gunthercox:
Looks good :+1: 

Thank you!

PR #69, opened 2015-10-25
Correct issue with corpus loading process

gunthercox:
This makes corrections to the loading process for loading corpus modules where an import error is occurring.


chatterbot/utils/module_loading.py
```diff
@@ -1,4 +1,5 @@
-import importlib
+from importlib import import_module as import_string
+from inspect import isclass
```
kevin-brown: Is this being used anywhere?

gunthercox: I don't believe so. I will make sure to remove them before this is merged. I'm currently getting better, results with these changes (basically attempting to use python modules for data instead of json files) but I'm still hitting an issue where importlib is returning a different sub-module depending on how it is called (works in tests, fails in application).

kevin-brown: If you aren't hitting the issue on Python 3, it might be an issue with relative imports vs absolute imports.

gunthercox: It appears to be both python ~2.7 and 3. I am planning on scrapping importing data files as python modules in favor of importing them as regular files. The advantage is that it is easier to traverse a directory and find all of the contained .json files than it is to traverse a python module to find all of the sub-modules.

PR #70, opened 2015-10-26
Commented the print statement in terminal.py io-adapter

anuragmishracse:
I have commented the "print" statement in terminal.py in adapters/io/,
because,
chatterbot.ChatBot.get_response("Some query"), would return a response as well as print that. It was getting printed because of the print statement in terminal.py. Removing that solves the issue.

### DIFF
--- a/chatterbot/adapters/io/terminal.py
+++ b/chatterbot/adapters/io/terminal.py
@@ -12,6 +12,6 @@ class TerminalAdapter(IOAdapter):
         return user_input
 
     def process_response(self, statement):
-        print(statement.text)
+#        print(statement.text)
         return statement.text
 


gunthercox:
I believe you might be looking to use the [`no_output`](https://github.com/gunthercox/ChatterBot/wiki/IO-adapters#no-output-adapter) adapter here instead of the `terminal` adapter.

You can switch the adapter by setting the `io_adapter` parameter:
```python
chatbot = ChatBot(
    "My ChatterBot",
    io_adapter="chatterbot.adapters.io.NoOutputAdapter"
)
```
https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/adapters/io/no_output.py

Please let me know if this solution works for you.

PR #77, opened 2015-11-09
Correction in matching svo

pradeephrish:
Updated index to 2


DarkmatterVale:
Hmm....It appears this is a little bit of a larger problem than just the index. I forgot that my library is only usable for Python 2.7 (I plan on adding 3.x support in the near future), and so it is throwing errors.

To fix this, I will remove regex4dummies as a library, and instead just use the pattern library. This will require some code change of the logic adapter.

Thank you for checking this out! I can't believe I missed it!

DarkmatterVale:
We are moving the commits over to a new branch where we are going to work on the adapter. For that reason, I am going to close this pull request.

Thank you for bringing the issue to my attention.

PR #80, opened 2015-11-23
Add evaluate_mathematically

DarkmatterVale:
Here is my first attempt at adding a way to evaluate mathematical expressions. @gunthercox what do you think? What are your thoughts on improving implementation?

A couple of notes. This is not finished, but I am at a point where I need to get input on how to implement it into the program. It currently looks at every incoming question or statement and attempts to decide whether or not the point is to evaluate a mathematical expression. This sort of works, but you run into issues pretty easily once you start describing the quantity of an object. It might think you are trying to have a mathematical expression evaluated, and then it will give the incorrect answer.

Another thing to note is the way I structured the function. I decided to create a new kind of adapter because this is definitely not an io adapter, nor is it a storage adapter. You could argue it is a logic adapter, but it doesn't work independent of another logic adapter.

Also, the reason the build is failing is because the answer in Python 3.x comes out to ```1.0``` while in 2.7 it is ```1```.


gunthercox:
@DarkmatterVale This is awesome. I will go over this pull request and provide some thorough feedback as soon as possible.

So far I think I like the concept of the processor adapter. This might be a really good addition in the future for determining if logic adapters need to run in other multi-logic adapter scenarios.

DarkmatterVale:
I had a similar thought process and agree that this could be very useful in the future. This will come especially in handy when you need a logic adapter in addition to a context adapter (or other kinds of adapters).

DarkmatterVale:
I have renamed the addition so that evaluate_mathematically() is now a plugin. This is a much better name for the function's purpose.

Also, plugins should allow for "external" functionality to be implemented, while Logic adapters should act as the way the user interfaces with the chatbot. In the future, we can use the plugins directory to add new functions such as sending emails, reading twitter messages, etc.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+    def isInteger(self, string):
+        """
+        If the string is an integer, returns
+        the int of the string. Otherwise,
+        it returns False.
+        """
+
+        if string.isdigit():
```
gunthercox: Will this handle negative integers?

If not, it should be possible to detect them like this:

```python
string.lstrip('-').isdigit()
```

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def isFloat(self, string):
+        """
+        If the string is a float, returns
+        the float of the string. Otherwise,
+        it returns False.
+        """
+
+        try:
+            float( integer )
```
gunthercox: It looks like `integer` is undefined here.

Also, this can probably be made to handle the detection of floats without a try, except block:
```python
string.lstrip('-').replace('.','',1).isdigit() # Handles detection of positive and negative floats
```

chatterbot/chatterbot.py
```diff
@@ -80,6 +83,14 @@ def get_response(self, input_text):
         """
         input_statement = Statement(input_text)
 
+        # Applying plugin logic to see whether the chatbot should respond in this way
+        plugin_response = self.plugin_chooser.choose( input_statement )
+
+        if plugin_response == False:
```
gunthercox: I think it might be possible to simplify this to:

```python
if not plugin_response == False:
    return plugin_response
```

tests/base_case.py
```diff
@@ -70,4 +70,3 @@ def setUp(self):
         self.chatbot.train(data1)
         self.chatbot.train(data2)
         self.chatbot.train(data3)
-
```
gunthercox: Not a major issue, but PEP8 does specify [a blank](http://pep8.readthedocs.org/en/latest/intro.html) line at the end of files.

chatterbot/adapters/plugins/evaluate_mathematically.py
kevin-brown: I honestly think the `try...except` block is a bit more clear here, and it avoids any edge cases with determining what is and isn't a float. If Python thinks it's a float, why argue with it?

You may want to try to coerce this to a `Decimal` though, that way you don't need to worry about floating point oddities.

```python
try:
    return decimal.Decimal(string)
except decimal.DecimalException:
    return False
```

If you want to be _really careful_, there are [other edge cases](https://github.com/tomchristie/django-rest-framework/blob/ff0eeaf77addc29d1ce0c10f47948ad06611f526/rest_framework/fields.py#L924-L936) that we handle in DRF for parsing decimal values.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Evaluates a set of expressions
+        and produces an answer. Then,
+        it returns the answer.
+        """
+
+        return eval( expression )
+
+
+    def isFloat(self, string):
```
kevin-brown: Just a friendly note about [PEP8](https://www.python.org/dev/peps/pep-0008/), the Python style guide, along with consistency within a code base.

For some reason these three methods are in `camelCase`, while Python is traditionally in `snake_case`.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,200 @@
...
+        # Returning normalized text
+        return string
+
+    def load_data( self, language ):
+        """
+        Load language-specific data
+        """
+
+        if language == "english":
+            with open(os.path.join(os.path.dirname(__file__), 'data') + "/math_words_EN.json") as data_file:
```
kevin-brown: Instead of forcing `/` as the directory separator here, you can add `math_words_EN.json` as a third parameter to the `os.path.join` call which is already being used here.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+
+        expression = []
+        string = ''
+
+        for chunk in input_text.split( ' ' ):
+
+            is_integer = self.isInteger( chunk )
+
+            if is_integer == False:
```
kevin-brown: In Python you should use `if x is False` to check for `False` instead of using `==`.

```python
>>> 0 == False
True
>>> 0 is False
False
```

DarkmatterVale: Thank you for this reminder! I have fixed this in the next commit.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: Sorry about that...I have learned camelCase in all my Java classes ( I'm still in high school and am currently taking AP CS and so I have previously worked with different styles )

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: Sounds like a great idea to coerce it into a Decimal, but I don't think we need to be that careful...If it appears to be really needed I can add that in later.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: Yes, I have made sure that it will handle negative integers without any issues

tests/base_case.py
DarkmatterVale: I don't know why this is happening because Atom shows a blank line at the end of the file, but I will see if I can fix this in a future commit

tests/test_chatbot_output.py
```diff
@@ -207,4 +224,3 @@ def test_database_is_not_updated_when_read_only(self):
 
         self.assertFalse(exists_before)
         self.assertFalse(exists_after)
-
```
DarkmatterVale: I'm not sure why this is happening with the last line of a file...I will look at how to fix this with Atom within the next week and incorporate additional lines into a future commit.

kevin-brown: This looks like Atom was removing the extra blank line at the end of the file.

There were two blank lines before, now there is only one.

DarkmatterVale: That makes more sense than my theory :)

chatterbot/adapters/plugins/evaluate_mathematically.py
kevin-brown: We are lacking tests for negative numbers, which is why I'm also curious about how they are being handled.

`isdigit` will return `False` if there are any non-digits (including decimals and the negative sign) in the string.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+    def is_operator(self, string):
+        """
+        If the string is an operator, returns
+        said operator. Otherwise, it returns
+        false.
+        """
+
+        if string in "+-/*^\(\)":
```
kevin-brown: Is there any reason why you appear to be escaping the parentheses `()` in this string? These shouldn't need to be escaped, and right now it's causing `\` to be detected as a valid operator.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+        Normalizes input text, reducing errors
+        and improper calculations.
+        """
+
+        # Setting all words to lowercase
+        string = string.lower()
+
+        # Removing punctuation
+        if string.endswith( ('.', '!', '?', ':', ';' ) ):
+            string = string[ : len(string) - 1 ]
```
kevin-brown: You can get everything but the last character with `string[:-1]` (or `string[0:-1]` if you want to be slightly more explicit).

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+
+        condensed_string = condensed_string.split( '_' )
+        for chunk_index in range( 0, len( condensed_string ) ):
+            value = ""
+
+            try:
+                value = str( eval( condensed_string[ chunk_index ] ) )
+
+                condensed_string[ chunk_index ] = value
+            except:
```
kevin-brown: What exceptions might happen here that you are trying to catch?

It's generally recommended to [avoid Pokemon exception handling](http://stackoverflow.com/q/10594113/359284), especially when you know what errors _should_ be raised.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+    def substitute_words(self, string):
+        """
+        Substitutes numbers for words.
+        """
+
+        self.load_data( "english" )
+
+        condensed_string = '_'.join( string.split( ' ' ) )
```
kevin-brown: I'm reading on through this code, and I can't figure out the reason for handling all spaces as underscores.

My guess would be that it has something to do with the use of regular expressions.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: I am catching all exceptions that rise from attempting to evaluate a non-numerical statement. Because I go through all chunks (including those that are not math expressions) I need to catch all exceptions that might be thrown.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: Yes, I am not sure why I did that. I have fixed it.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: I added tests in my next commit. They were previously being treated as floats, which I have now changed by replacing the is_integer() code with:

```
try:
            return int( string )
except:
            return False
```

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: When I generate numbers out of words, I create spaces. This makes it really difficult to attempt to separate the code based on spaces because I have just created additional spaces...

I chose the underscore character because it is not commonly used in English. Occasionally it will be used, but not in most conversations.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: I thought that as well, but when I looked it up (to make sure I was doing everything correct), I found that using ```endswith()``` is a better route to go.

DarkmatterVale:
@gunthercox and @kevin-brown Do you guys have any more questions? Do I need to give a more in-depth explanation for how the code works?

gunthercox:
I'm checking this out locally using `examples/terminal_example.py`. It appears that when I enter a value that contains a mathematical expression, such as "What is 5 + 5?", no response is returned. This is strange since I can clearly see that tests are passing.

DarkmatterVale:
I had the same problem. In fact, I am unable to see any changes I make to the code when I run the examples. The best way to test, which is the way I do it because of this issue, is to add a print statement right before the return statement. The before and after code is below.

Before:
```
if not plugin_response is False:
     return plugin_response
```

After:
```
if not plugin_response is False:
     print plugin_response
     return plugin_response
```

gunthercox:
In general, a good way to test out changes to python packages that are also imported in other local projects is just to install the package (with the changes you have made) locally.

This can be done using:
```
pip install . --upgrade
```

This should be run from within the same directory as `setup.py`.

***
Also, another item I noticed is that you will need to include the json data file you created in the installed package. This can be done by adding the following line to `MANIFEST.in`:
```
recursive-include chatterbot/adapters/* *.json
```

DarkmatterVale:
Thank you for that information! I will not make these mistakes again; sorry for not doing things properly in the first place!

I was able to test ( correctly this time ) the adapter and I found the problem. I was not processing the response properly, leading it to not being displayed. This is fixed in the next commit along with the ```MANIFEST.in``` update.

gunthercox:
No problem, this is all really good work and its cool to see it coming together. Also, that corrected the issue.

DarkmatterVale:
Awesome, glad that fixed the issue! I will remember all of this next time I make changes so you guys don't have to spend hours reviewing code :D

How does it work? I know it doesn't work for all math expressions ( far from it ), but it is a start and a place to work from.

gunthercox:
Code review is a part of programming. It's always possible to make a mistake, having a few more sets of eyes on it never hurts.

So far everything that I'm seeing here looks great. Performance-wise it works excellently. At the moment I believe there is only one more thing in the code that I believe could be improved, I'll leave a comment about it on the line.

chatterbot/adapters/plugins/evaluate_mathematically.py
gunthercox: Just a note that might be useful here, calling `string.split(' ')` will split the string at every space character. It might be a good idea to use `string.split()` instead. If no character is passed in, the split method will split the string for any number of white space characters (including multiple consecutive spaces as well as newline and tabs which might just so happen to fall into the input string).

DarkmatterVale: Ok; that is a much better way to remove whitespace. I have fixed it and am waiting to push the change to see if you guys have any other comments...I don't like making messy git histories

gunthercox: Sure, I will run through again and do another sweep. I might not get a chance to do it until later this week. I'll let you know if I find anything.

DarkmatterVale: Thank you!

I will also do a check to see if I can find any more errors.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+    def simplify_chunks(self, input_text):
+        """
+        Separates the incoming text.
+        """
+
+        string = ''
+
+        for chunk in input_text.split( ' ' ):
```
DarkmatterVale: Should I use ```input_text.split()``` instead of ```input_text.split( ' ' )```? That way I would be separating the words based on all whitespace characters not just spaces.

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+            if is_chunk_integer is False:
+                is_chunk_float = self.is_float( chunk )
+
+                if is_chunk_float is False:
+                    is_chunk_operator = self.is_operator( chunk )
+
+                    if is_chunk_operator is False:
+                        continue
+                    else:
+                        string += str( is_chunk_operator ) + ' '
```
DarkmatterVale: This could be simplified to:
```
if not is_chunk_operator is False:
     string += str( is_chunk_operator ) + ' '
```

chatterbot/adapters/plugins/evaluate_mathematically.py
gunthercox: Yup, `input_text.split()` is the way to go here :+1:

chatterbot/adapters/plugins/evaluate_mathematically.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Normalizes input text, reducing errors
+        and improper calculations.
+        """
+
+        # Setting all words to lowercase
+        string = string.lower()
+
+        # Removing punctuation
+        if string.endswith( ('.', '!', '?', ':', ';' ) ):
```
gunthercox: If might be a good idea to handle any non-alphanumeric character rather than just this specific set.
`if not string[-1].isalnum():` would check if the last character in the string is not an alphanumeric character.

chatterbot/adapters/plugins/evaluate_mathematically.py
gunthercox: @DarkmatterVale I agree with kevin on this.

chatterbot/adapters/plugins/evaluate_mathematically.py
DarkmatterVale: That is a good idea & I have updated the code in the latest commit.

gunthercox:
@DarkmatterVale These changes look good to me. Did you have any more additions before I merge this?

DarkmatterVale:
No I do not have any additional changes to make.

gunthercox:
Alright, just did a manual merge with master because there was some conflicts that GitHub couldn't resolve automatically. Either way, this is an awesome addition and it will be available in the next release on pypi.

PR #87, opened 2015-11-29
ClosestMeaningAdapter wasn't useable

bloodywing:
TypeError: __init__() got an unexpected keyword argument 'logic_adapter because chatterbot.py initializes the logic adapters with his kwargs. The Closestmatch adapter uses the __init__() from logic.py


gunthercox:
Thank you. Can't believe I missed this! :+1:

PR #88, opened 2015-12-01
Add Tkinter example GUI

DarkmatterVale:
I created a simple GUI that users can run to interact with ChatterBot.


gunthercox:
This looks great :+1:

PR #91, opened 2015-12-06
Restructure logic adapters to return response

gunthercox:
Currently, logic adapters select a match based on a given algorithm and return that match. This pull request makes changes so that they will return the response instead of the match. The reason this is needed is because as ChatterBot continues to develop, it is likely that there will be logic adapters in the future that do not always follow this pattern, thus it would be better if the current selection of logic adapters completely encompassed all of the logic they require to produce a response.

- [x] Reconfigure logic adapters to have access to storage adapters
- [x] Move response selection logic inside of logic adapters
- [x] Do something with those three response selection methods in the main class


DarkmatterVale:
This is a much needed update. Thank you for working on this!

gunthercox:
No problem.

chatterbot/adapters/adaptation.py
```diff
@@ -0,0 +1,35 @@
...
+class Adaptation(object):
+    """
+    This is a base from which ChatterBot inherits utility methods
+    and a context attribute that allows access of adapters to be
+    shared between other adapters. This also makes it possible to
+    share other context information such as a name, or the current
+    conversation with each of the adapters.
+    """
+
+    class context(object):
```
kevin-brown: Any particular reason why you have this set up as a static class with a single static property?

gunthercox: The main reason this is static is because it needs to hold the variable for the current conversation. Having it be static made more sense than passing the current conversation to each adapter every time a method on the adapter that needs that information is called.

gunthercox: Actually, there might be a problem with this. I had thought that static classes in python could be updated with out effecting each other _if_ they were inside on an instance of another class. So for example...

```python
class A:
    class B:
        x = 0

# then ...

a = A()
b = A()

a.B.x = 1
b.B.x = 2

print a.B.x
>>> 1
print b.B.x
>>> 2
```

However this does not appear to be the case. It looks like the subclass (B) is actually updated across every instance of A. Do you know if it would be possible to keep class B from changing other instances of A?

kevin-brown: Because both of these are static, any changes made to the class of the `data` property within it will be made across all instances. If you are looking for the context to be instance-specific, you need to create a new instance of the class and assign it as a attribute during initialization.

```python
class B:
    def __init__(self):
        self.x = 0

class A:
    def __init__(self):
        self.b = B()
```

```python
>>> a = A()
>>> b = A()
>>> print(a.b.x)
0
>>> print(b.b.x)
0
>>> a.b.x = 1
>>> b.b.x = 2
>>> print(a.b.x)
1
>>> print(b.b.x)
2
```

Note that if you made the `x` attribute static as well, you would face a similar issue within the nested class.

PR #99, opened 2015-12-20
New remote MongoDB adapter.

sgoblin:
The old adapter for MongoDB could not connect to remote databases, so I forked it into a new adapter and added code to make work with remote DBs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ settings.py
 *.pyc
 *.swp
 *.egg-info
+.eggs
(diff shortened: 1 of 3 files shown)


sgoblin:
I'm not sure how to add tests for this, as it would require setting up a remote MongoDB. I could set up a MongoLab DB for this, but @gunthercox would have to add the password to the .travis.yml file as encrypted.

kevin-brown:
This would probably be better integrated into the existing MongoDB adapter, with the database location being a configuration option (defaulting to localhost).

sgoblin:
Actually, I think you're right. How do you add defaults for **kwargs?

gunthercox:
@sgoblin You can add a default for kwargs like this:

```python
something = kwargs.get("some_key", "default_value")
```

PR #100, opened 2015-12-25
Added remote mongodb capabilities to adapter.

sgoblin:
Added kwarg for database_uri, which is in the standard MongoDB URI format.

This is a rewrite of PR #99 , where instead of a new adapter, the feature is just written into the new adapter.

### DIFF
--- a/chatterbot/adapters/storage/mongodb.py
+++ b/chatterbot/adapters/storage/mongodb.py
@@ -106,7 +106,7 @@ class MongoDatabaseAdapter(StorageAdapter):
             data = statement.serialize()
 
             # Remove the text key from the data
-            self.statements.update({'text': statement.text}, data, True)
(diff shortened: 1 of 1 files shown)


chatterbot/adapters/storage/mongodb.py
```diff
@@ -105,7 +106,7 @@ def update(self, statement):
             data = statement.serialize()
 
             # Remove the text key from the data
-            self.statements.replace_one({'text': statement.text}, data, True)
+            self.statements.update({'text': statement.text}, data, True)
```
gunthercox: I don't believe this was meant to be changed, `replace_one` is correct.

sgoblin:
@gunthercox I fixed the thing, don't know how that got in there in the first place! Guess I should check my diffs before committing, eh?

gunthercox:
Looks great! Thank you.

PR #103, opened 2015-12-30
Add logic adapter that returns the current time when asked

gunthercox:


DarkmatterVale:
Since we are using logic adapters for many of these additional functions, should we remove plugin adapters? Or are there still functions that we should implement as plugins?

I would believe that functions that are deeply integrated into the chatbot, such as time, date, location, etc, should be added as logic adapters while functions that are "extra", or not core to chatterbot be added as plugins (which are essentially logic adapters but more complex). What do you think @gunthercox ?

gunthercox:
@DarkmatterVale Logic adapters and plugin adapters both accomplish the same task. I think it would be beneficial to make a move to migrate the existing selection of plugin adapters to become logic adapters (this should not require a huge amount of modification). Apart from sharing a more unified code base, the other benefit is that there will be no restrictions to define what parts of the program are "deeply integrated" and what parts are not.

chatterbot/adapters/logic/time_adapter.py
```diff
@@ -0,0 +1,32 @@
...
+            ("do you know the time", 1),
+            ("do you know what time it is", 1),
+            ("what is the time", 1),
+            ("do you know the time", 0),
+            ("it is time to go to sleep", 0),
+            ("what is your favorite color", 0),
+            ("i had a great time", 0)
+        ]
+
+        self.classifier = NaiveBayesClassifier(training_data)
```
DarkmatterVale: We should separate classifiers from the logic adapters that use them. By doing this, it will give a simpler and more flexible structure for logic adapters.

gunthercox: I'm not sure I understand what you mean. Are you referring to the the training data or the NaiveBayesClassifier class?

DarkmatterVale: Sorry for the confusing message. I am referring to the NaiveBayesClassifier. I would create different classifiers, such as a NaiveBayesClassifier, a neural network classifier, etc, and then call those from within the logic adapters. This allows for flexibility & ease of updating if, for example, you have a lot of logic adapters that use the same code. If that code needs an update, it would be simpler and cleaner to just have to update the code in a single location rather than many locations.

gunthercox: Ok, I agree with that. Let me give this some thought because I have been considering doing something similar with parts of the logic in the closest_meaning and closest_match adapters. I hand not been thinking in terms of classifiers so that may be a good approach to take. The only constraint or possible road block I can possibly see is the case that each classifier cannot be made generic enough to be usable in other existing or new adapters.

Again, I'll need to give this some thought and experiment a bit with the concept, but it might be a step in the right direction. I many end up opening a new ticket about this if I come up with a good solution.

DarkmatterVale: Any new thoughts? How has experimentation gone?

gunthercox: Sorry, haven't had a chance yet.

DarkmatterVale: Ok, no problem! I just wanted to hear about any great ideas you had :D

gunthercox: I think for now I am going to merge this. I may work on possibly splitting up the logic more in a future pull request.

DarkmatterVale: Ok. I can also work on that since I'm just about to be done with GCI.

Issue #111, opened 2016-01-25
Logic adapters not processing correctly


DarkmatterVale:
Although it looks fine on the surface, there are issues with the current method for using multiple logic adapters. This can be easily found when you run chatterbot and type "what time is it?". Sometimes, I will get the current time while other times I will get "what time is it?", which is displayed because there is no other statement in the database. In addition, most of the time when I ask that question later in the conversation, I do not get the actual time but a response from the database.

After debugging the error for a fair amount of time, I found where the problem is coming from. The current way responses are being prioritized makes getting the correct logic adapter to respond very difficult. I am not sure why this is occurring, but the problem is in https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/adapters/logic/multi_adapter.py#L11. You can test this by adding a print statement in evaluate_mathematically in the try block (located at https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/adapters/logic/evaluate_mathematically.py#L31) printing the expression after assigning it the evaluated input. If you then type in something like "what is 100 * 10?", you will see that evaluate_mathematically properly generates a response, and returns a confidence of 1, but still the result is not used.

Further, the database responses generated from the evaluate_mathematically & time functions are incorrect. This might be unrelated, or the program might be checking the database before checking the other adapters first, and when it gets a 1 confidence for an equal statement it returns that statement and doesn't replace it with the generated result from other logic adapters. This is an issue particularly for logic adapters that would have been considered plugins, which do not want their results saved to the database.

To test, the order of my logic adapters is as follows:
"    logic_adapters=[
        "chatterbot.adapters.logic.EvaluateMathematically",
        "chatterbot.adapters.logic.TimeLogicAdapter",
        "chatterbot.adapters.logic.ClosestMatchAdapter"
    ],"

@gunthercox Any ideas how to fix this? I have 2 separate solutions that I have been able to come up with so far:

1) Re-implement plugin support. By classifying them differently, we can easily fix the database issue as well as put higher priority on getting a result from a plugin, which will fix the issue regarding incorrect order.

2) Leave the system as-is, but change the way adapters are processed. This will make the code less readable, but may be a much better solution than 1. It would require adding a flag to tell chatterbot whether to save the result and the statement to the database. In addition, it would require a new way to process and give confidence values for logic adapters. I would argue that we remove the tie breaker, and use the following logic instead (the for loop just stays the same):
[...]

PR #112, opened 2016-01-25
minor fixes in logic adapters & terminal example

DarkmatterVale:
Just a couple of minor changes, including:

1) Add additional logic adapters to the terminal example adding functionality
2) Possible bug fix in evaluate_mathematically (not really sure, but this makes sure no bug will ever arise)
3) Sometimes, the time logic adapter would find a statement similar to "what is 100 * 10?" as something it should process. By adding the additional piece of training data, the time adapter does not incorrectly process math statements as often

### DIFF
--- a/chatterbot/adapters/logic/evaluate_mathematically.py
+++ b/chatterbot/adapters/logic/evaluate_mathematically.py
@@ -25,11 +25,11 @@ class EvaluateMathematically(LogicAdapter):
         input_text = statement.text
 
         # Getting the mathematical terms within the input statement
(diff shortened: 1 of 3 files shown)


gunthercox:
:+1: Changes look good.

Also, I'm currently looking into #111. Should have some sort of feedback shortly.

DarkmatterVale:
Awesome! Thank you!

Also, on a side note, GCI ended today, so I hope to be able to spend some more time working on chatterbot now. Thank you!

PR #113, opened 2016-01-25
Make sure that closest match adapter confidence returns value between 0 and 1

gunthercox:
This updates and adds tests to ensure that the confidence value returned by the closest match adapter fall within the range between 0 and 1.

### DIFF
--- a/chatterbot/adapters/logic/closest_match.py
+++ b/chatterbot/adapters/logic/closest_match.py
@@ -35,6 +35,9 @@ class ClosestMatchAdapter(BaseMatchAdapter):
             limit=1
         )[0]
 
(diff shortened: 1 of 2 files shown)


gunthercox:
@DarkmatterVale I believe this corrects the issue observed in #111. Feel free to review and merge this pull request.

DarkmatterVale:
@gunthercox Did it fix the problem for you? Can you now accurately get a response from the proper adapter?

Regarding the issue with the database, we can discuss that in the issue and not this PR.

gunthercox:
This did correct the issue.

DarkmatterVale:
Awesome, thank you for fixing it! I'm gonna go ahead and merge it in then.

PR #114, opened 2016-01-26
Separate classifier out

DarkmatterVale:
This PR separates the classifier out from the logic adapter. Although the current NaiveBayes classifier is pretty simple, future classifiers might be much more complicated. Because of this, separating the classifiers out will make the code much easier to understand in the future.

### DIFF
--- a/chatterbot/adapters/logic/__init__.py
+++ b/chatterbot/adapters/logic/__init__.py
@@ -5,3 +5,4 @@ from .time_adapter import TimeLogicAdapter
 from .multi_adapter import MultiLogicAdapter
 from .no_knowledge_adapter import NoKnowledgeAdapter
 from .evaluate_mathematically import EvaluateMathematically
+from .classifier_naivebayes import NaiveBayes_Classifier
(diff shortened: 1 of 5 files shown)


DarkmatterVale:
@gunthercox I know you had different ideas about this. Feel free to review and try it out.

gunthercox:
I'm not sure if classifiers would make sense as a subclass of Adapter since I don't believe they will be used the same was as each of the three adapter types. Perhaps just creating a Classifier base class with a namespace at `chatterbot.classifiers` would be a better direction to go in.

Still... at this point, I am still not sure that I see the need for a separate set of classifier objects. It is important to consider the fact that many machine learning algorithms (such as Naive Bayes Classification) are relatively specific to the types of problems they can solve. And in many cases, it is better to have a machine learning algorithm that is optimized for a specific problem, rather than a generic algorithm designed to solve a range of problems.

Some things to think about:
1. What other classifiers are needed / would be useful?
2. What is the [need](http://www.gettingagile.com/2006/09/25/need-driven-design-as-an-integration-strategy/) that is driving this addition?
3. How do you define a classifier and what does it do? This is important for designing a good interface for the base class.
4. Is there any existing code (aside from the time adapter) that contains logic that would work as a classifier?

DarkmatterVale:
@gunthercox I agree with you, and I think at the moment this addition is not needed.

I will close the PR.

PR #115, opened 2016-01-26
Add custom implementations for various NLTK functions

DarkmatterVale:
This PR includes simple implementations of the NLTK functions used by the closest meaning adapter. The structure of the system makes it easy to improve each individual component very easily when we want to make the analysis more complex.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,6 +4,9 @@ from unittest import TestCase
 from chatterbot.utils.clean import clean_whitespace
 from chatterbot.utils.clean import clean
 from chatterbot.utils.module_loading import import_module
+from chatterbot.utils.pos_tagger import POSTagger
+from chatterbot.utils.stop_words import StopWordsManager
+from chatterbot.utils.word_net import Wordnet
 
 
 class UtilityTests(TestCase):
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -12,6 +15,22 @@ def test_import_module(self):
...
+    def test_stop_words(self):
+        stopwords_manager = NLTKStopWordsManager()
+        words = stopwords_manager.words("english")
+        test_case = set(["too"]) - set(words)
+
+        self.assertEqual(test_case, set([]))
+
+    def test_word_net(self):
+        wordnet = NLTKWordnet()
+        synsets = wordnet.synsets('test')
```
gunthercox: The case for `test_word_net` was probably intended to assert something i'm guessing?

gunthercox:
I like the direction this is going in to break some of the code out into utils. What would your thoughts be on renaming these classes to be more generic? It is possible in the future that we may use something other than NLTK to accomplish these tasks.

NLTKWordnet --> Wordnet
NLTKStopWordsManager --> StopWordsManager
NLTKPOSTagger --> POSTagger (or PartOfSpeechTagger to be more verbose)

DarkmatterVale:
@gunthercox Yes, that is probably a good idea; I had the names including NLTK because I meant them as wrapper classes for NLTK's built-in functions, but that is not really needed. I have changed the names to match what you suggested.

gunthercox:
Awesome :+1: Looks good.

PR #116, opened 2016-01-27
Multiple io adapters

DarkmatterVale:
This PR add support for using multiple io adapters.

At the moment, I am currently just getting input from the first io adapter. To allow multiple inputs, we may have to restructure a couple of things. I haven't thought too much about it yet though, so I'm not sure what will have to be changed.


gunthercox:
It looks like there was a conflict after one of the other pull requests was merged. This will need to be rebased.

DarkmatterVale:
@gunthercox I rebased and fixed the conflicts.

gunthercox:
This looks great :+1: 

Would you mind updating the documentation in the wiki? You can probably do something very similar to what we have for logic adapters already: https://github.com/gunthercox/ChatterBot/wiki/Logic-adapters#using-multiple-logic-adapters

PR #117, opened 2016-01-27
Add a MacOSX TTS io adapter

DarkmatterVale:
I added an io adapter for a TTS engine built in to MacOSX.

Although I added an output adapter, I have not added a corresponding STT adapter so it gets input using the Terminal. I recommend we make a change to the way io adapters are structured (which will also solve the problem with multiple inputs in my other PR), separating out input adapters and output adapters. This way, the user can specify more precisely what they would like to use for inputs and outputs. @gunthercox What do you think?


The-Penultimate-Defenestrator:
This is awesome.

gunthercox:
This looks like a good addition. I don't have access to a mac to test this on, but I trust that it works.

PR #119, opened 2016-01-28
Add basic class documentation

DarkmatterVale:
Most of the documentation is pretty simple, but the structure of ChatterBot is pretty straightforward.

### DIFF
--- a/chatterbot/adapters/adapter.py
+++ b/chatterbot/adapters/adapter.py
@@ -1,4 +1,7 @@
 class Adapter(object):
+    """
+    An abstract superclass for all adapters
+    """
 
     def __init__(self, **kwargs):
         self.context = None
--- a/chatterbot/adapters/io/io_json.py
+++ b/chatterbot/adapters/io/io_json.py
@@ -3,6 +3,10 @@ from chatterbot.conversation import Statement
 
 
 class JsonAdapter(IOAdapter):
+    """
+    The JsonAdapter class returns a python dictionary which
+    is useful for serialization and deserialization of statements.
+    """
 
     def process_input(self, input_json):
         """
--- a/chatterbot/adapters/logic/closest_match.py
+++ b/chatterbot/adapters/logic/closest_match.py
@@ -4,6 +4,14 @@ from fuzzywuzzy import process
 
 
 class ClosestMatchAdapter(BaseMatchAdapter):
+    """
+    The ClosestMatchAdapter creates a response by
+    using fuzzywuzzy's process class to extract the most similar
+    response to the input. This adapter selects a response to an
+    input statement by selecting the closest known matching
+    statement based on the Levenshtein Distance between the text
+    of each statement.
+    """
 
     def get(self, input_statement, statement_list=None):
         """
--- a/chatterbot/adapters/logic/closest_meaning.py
+++ b/chatterbot/adapters/logic/closest_meaning.py
@@ -7,6 +7,14 @@ from chatterbot.utils.word_net import Wordnet
 
 
 class ClosestMeaningAdapter(BaseMatchAdapter):
+    """
+    This adapter selects a response by comparing the tokenized form of the
+    input statement's text, with the tokenized form of possible matching
+    statements. For each possible match, the sum of the Cartesian product of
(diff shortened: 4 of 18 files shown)


chatterbot/adapters/adapter.py
```diff
@@ -1,4 +1,7 @@
 class Adapter(object):
+    """
+    An abstract class all adapters are subclasses of.
```
gunthercox: Maybe reword this as
> An abstract subclass for all adapters.

(just to make it grammatically clear).

chatterbot/adapters/io/io_json.py
```diff
@@ -3,6 +3,10 @@
 
 
 class JsonAdapter(IOAdapter):
+    """
+    The JsonAdapter class is an IO adapter that allows
+    ChatterBot to interface with Json structured databases.
```
gunthercox: This is not true. The json adapter returns data in a python dictionary format which is useful for serialization and deserialization of statements. One place this could be used may be when creating a web based api. It is unlikely that this adapter would be used with a database directly.

chatterbot/adapters/logic/closest_match.py
```diff
@@ -4,6 +4,10 @@
 
 
 class ClosestMatchAdapter(BaseMatchAdapter):
+    """
+    The ClosestMatchAdapter extends the BaseMatchAdapter to find
+    the closest matching response.
```
gunthercox: This docstring doesn't actually add any information to this class. It just describes what the first line of the class is doing.

DarkmatterVale:
@gunthercox I have updated the docstrings; do they look better now?

chatterbot/adapters/logic/mixins.py
```diff
@@ -1,4 +1,9 @@
 class TieBreaking(object):
+    """
+    TieBreaking determines which adapter's result should be used
+    in the event that mutliple adapters return the same highest
+    confidence value.
```
gunthercox: Tie breaking isn't actually used in for tie breaking between multiple adapters. If two adapters return the same confidence value, then the adapter that is listed first will take priority. The TieBreaking class is used in a single adapter when there are multiple known statements that are exact matches for a given input statement.

chatterbot/adapters/logic/closest_meaning.py
```diff
@@ -7,6 +7,10 @@
 
 
 class ClosestMeaningAdapter(BaseMatchAdapter):
+    """
+    The ClosestMeaningAdapter extends the BaseMatchAdapter to
+    find the response with the closest meaning.
```
gunthercox: Pretty much any programmer will read `ClosestMeaningAdapter(BaseMatchAdapter)` as "The ClosestMeaningAdapter extends the BaseMatchAdapter" and based on it's name they will probably assume that it "find the response with the closest meaning".

Maybe we can provide some information that will help someone understand how the logic in this class generally works. In this case I might describe how this class functions as:

> This adapter selects a response by comparing the tokenized form of the input statement's text, with the tokenized form of possible matching statements. For each possible match, the sum of the Cartesian product of the path similarity of each statement is compared. This process simulates an evaluation of the closeness of synonyms. The known statement with the greatest path similarity is then returned.

chatterbot/adapters/logic/closest_match.py
```diff
@@ -4,6 +4,12 @@
 
 
 class ClosestMatchAdapter(BaseMatchAdapter):
+    """
+    The ClosestMatchAdapter extends the BaseMatchAdapter to find
+    the closest matching response. It creates a response by
+    using fuzzywuzzy's process class to extract the most similar
+    response to the input.
```
gunthercox: Maybe just leave out the "The ClosestMatchAdapter extends the BaseMatchAdapter to find the closest matching response."?

Also, instead of implying that fuzzywuzzy just magically chooses a response, perhaps it would be more helpful to say something like
> This adapter selects a response to an input statement by selecting the closest known matching statement based on the Levenshtein Distance between the text of each statement.

DarkmatterVale:
@gunthercox Do you have any more comments? If not, can we merge this in?

gunthercox:
This looks good :+1:

PR #121, opened 2016-01-31
Add a simple weather logic adapter

DarkmatterVale:
This PR adds a simple weather forecast logic adapter. There are a couple of issues, however, including:

1) Warnings that need to be "squashed"
2) Ideally, we remove using python-forecastio and create our own implementation, but that is work for the next version (an enhancement and not the initial version)

Also, since it requires an API key, we cannot create tests for the adapter.


gunthercox:
:ghost: Looks good... merging.

PR #122, opened 2016-02-01
Add a developer assistant

DarkmatterVale:
This PR adds a simple developer assistant. A lot more work is needed to give it flexible command understanding, but it can currently run python programs located on the user's computer with ease.

### DIFF
--- a/chatterbot/adapters/logic/__init__.py
+++ b/chatterbot/adapters/logic/__init__.py
@@ -1,7 +1,13 @@
+# Base logic adapters
 from .logic import LogicAdapter
-from .closest_match import ClosestMatchAdapter
-from .closest_meaning import ClosestMeaningAdapter
-from .time_adapter import TimeLogicAdapter
 from .multi_adapter import MultiLogicAdapter
 from .no_knowledge_adapter import NoKnowledgeAdapter
+
+# Match adapters
+from .closest_match import ClosestMatchAdapter
+from .closest_meaning import ClosestMeaningAdapter
+
+# Other adapters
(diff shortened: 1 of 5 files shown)


DarkmatterVale:
I will add tests and update the PR.

DarkmatterVale:
@gunthercox I have made a number of large updates to the adapter today, and I think it is ready to be added in. If you would like to test the final product, you can do so by running the following command (inside a chatbot instance):

run path_to_your_python_program.py

There are a couple of constraints so far, however, including:
1) Only python programs work (and there is no auto-language-detect)
2) You must use the command "run", although it can be used in multiple formats

An example of the output (after running the file once before this test):

Bot:    Type something to begin...
Me:     run main.py
Bot:    Would you like to use the path /Users/vtolpegin/github/regex4dummies/examples/accuracy_tester/?
Me:     yes

DarkmatterVale:
After some thought, I do not believe this is consistent with the current ChatterBot, as it is not a chat adapter but a developer assistant.

gunthercox:
@DarkmatterVale I agree, perhaps it would be possible to package the developer assistant as it's own library. I had been thinking of creating something similar at one point, my idea was to create an agent that would automatically detect issues with code on github and leave a comment or create a pull request.

DarkmatterVale:
@gunthercox I will create a new repository and put the code in there. I have also thought about creating something similar to this extensively, although my plans are a little more ambitious than yours...

DarkmatterVale:
You should be able to find the current project at https://github.com/DarkmatterVale/DevAssist

PR #125, opened 2016-02-03
Database update

DarkmatterVale:
I have changed the database configuration to be better suited to future logic adapters' needs. The changes include:

1) Add a database_dir variable: This specifies the directory for the database as well as other data that is created when using ChatterBot
2) Update the database configuration in jsondatabase.py to use the new format

This is a much needed change for new logic adapters which require data files to be created and edited.

### DIFF
--- a/chatterbot/adapters/storage/jsondatabase.py
+++ b/chatterbot/adapters/storage/jsondatabase.py
@@ -9,7 +9,7 @@ class JsonDatabaseAdapter(StorageAdapter):
     def __init__(self, **kwargs):
         super(JsonDatabaseAdapter, self).__init__(**kwargs)
 
-        database_path = self.kwargs.get("database", "database.db")
+        database_path = self.kwargs.get("database_dir") + self.kwargs.get("database", "database.db")
         self.database = Database(database_path)
 
     def _keys(self):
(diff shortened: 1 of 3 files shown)


gunthercox:
I'm not sure I see why this change is needed. Isn't is possible to get the directory the database is in from the database path variable?

Additionally, a file path is an attribute that does not apply to every type of database. Mongo db is a good example of this.

DarkmatterVale:
That is true, but for a number of logic adapters additional files need to be created once they are used (dynamic training data is a great example). Because there are multiple additional files that need to be generated, a single directory to hold all of those files makes the system look cleaner.

I have updated the PR to remove the code that could conflict with the use of a Mongo db. With the update, ChatterBot has a common, easily set location to store files generated by ChatterBot.

gunthercox:
> That is true, but for a number of logic adapters additional files need to be created once they are used (dynamic training data is a great example).

I actually don't know what adapters you are referring to here, I don't think chatterbot has any logic adapters that create additional files during runtime.

DarkmatterVale:
At the moment, no adapters need additional files. In the near future though, I anticipate using multiple additional files for logic adapters' training (but this is a complicated addition and takes time).

Further, it is better that we implement this now as opposed to another contributor having to figure it out on the spot when sending a PR.

gunthercox:
I am going to veto this change for now. I am a strong supporter of only writing code that is necessary. Unused code can add unnecessary complications and decrease readability.

PR #154, opened 2016-04-26
Create curse_pt-BR.json

tiagotmc:
A few possible curses from the user and a polite response from the system.


gunthercox:
I like the idea of teaching ChatterBot polite ways to respond to various insults and curses. At the moment I'm hesitant to merge this because the bot can possibly respond from either the input or the response in some cases.

PR #163, opened 2016-05-08
Split io adaper module up into two new adapter modules.

### LINKED ISSUE #133: Rename no output adapter
I just noticed that there is an extra parameter for in the `process_input` method of the JSON io adapter. A change needs to be made so that all io adapters meet the same interface specification.

gunthercox:
Input and output adapters are now separate. This change makes it possible to write new adapters more easily, reduces some duplicate code between different io adapters, and makes ChatterBot's input and output configuration more flexible as a whole.

- [x] Split up io adapters into input adapters and output adapters
- [x] Combine and rename remaining adapters as needed
- [x] Only allow one input and one output adapter to be specified
- [x] Update examples as needed
- [x] Update documentation

Closes #133 
[Changes breaking backwards compatibility]


chatterbot/adapters/input/variable_input_type_adapter.py
```diff
@@ -0,0 +1,60 @@
...
+
+class VariableInputTypeAdapter(InputAdapter):
+
+    def __init__(self, **kwargs):
+        super(VariableInputTypeAdapter, self).__init__(**kwargs)
+
+    def detect_type(self, statement):
+        input_type = type(statement)
+
+        if input_type is Statement:
```
kevin-brown: I'd recommend using `isinstance` over `type() is` here, to allow for subclasses to be detected properly.

chatterbot/adapters/output/multi_adapter.py
```diff
@@ -0,0 +1,31 @@
...
+        super(MultiOutputAdapter, self).__init__(**kwargs)
+
+        self.adapters = []
+
+    def process_response(self, statement):
+        """
+        Takes an input value.
+        Returns an output value.
+        """
+        for i in range(1, len(self.adapters)):
```
kevin-brown: It looks like you want to skip the first adapter and run the rest...

```python
for adapter in self.adapters[1:]:
```

Issue #148, opened 2016-04-13
Avoiding repetitions


SamuelRiversMoore:
Hi,
I'm wondering, is there is any simple ways to prevent the bot from repeating itself too much?

Maybe : keeping a record of the two or three previous exchanges and returning an output that could be a bit less appropriate if the most appropriate has been returned too recently… What do you think?

Amazing work, btw! :+1:

gunthercox:
I agree with this. There is currently a `recent_statements` variable which I added so that this could eventually be accomplished. Currently the variable holds a list of the bot's responses for the current session. At some point I would love to make the change so that it avoids repeating the same words frequently.

cd-rum:
+1

SamuelRiversMoore:
Ok, maybe it's not the cleanest way to do but I have something. At least it's a start.
I modified two functions : `process()` in `base_match.py` and `get()` in `closest_match.py`

For `get()` : 

   ```
    def get(self, input_statement, statement_list=None):
        """
        Takes a statement string and a list of statement strings.
        Returns the closest matching statement from the list.
        """
        statement_list = self.get_available_statements(statement_list) # entries in json db

        if not statement_list:
            if self.has_storage_context:
                # Use a randomly picked statement
                return [(0, self.context.storage.get_random())]
            else:
                raise EmptyDatasetException

        # Get the text of each statement
        text_of_all_statements = []
        for statement in statement_list:
            text_of_all_statements.append(statement.text)

        # Make a list of matches ordered by confidence
        matches = process.extract(
            input_statement.text,
            text_of_all_statements
        )
        results = [( confidence/100.0, next((s for s in statement_list if s.text == match), None )) for match, confidence in matches]
        return results
```


For `process()` : 

   ```
 def process(self, input_statement):

        recent_statements = self.context.recent_statements[-10:] # get the ten last recent statements

        results = self.get(input_statement)

        for e in results :
            confidence = e[0]
            closest_match = e[1]

            # Save any updates made to the statement by the logic adapter
            self.context.storage.update(closest_match)

            # Get all statements that are in response to the closest match
            response_list = self.context.storage.filter(
                in_response_to__contains=closest_match.text
            )

            if response_list:
                response = self.break_tie(response_list, self.tie_breaking_method)
            else:
                response = self.context.storage.get_random()

            if response not in recent_statements:
                break # break the loop if the response has not been said in the ten last exchanges

        return confidence, response
```

gunthercox:
[...]

PR #170, opened 2016-05-15
Redesign recent_statements to hold the entire current conversation

### LINKED ISSUE #132: Redesign recent_statements to hold the entire current conversation
The updated representation should be a list of tuples:
```
[
    (input_statement, output_statement),
    (input_statement, output_statement),
    (input_statement, output_statement),
   ...
]
```
This change will be needed for upcoming additions of logic adapters that look at the context of the current conversation.

gunthercox:
Closes #132
Related to avoiding repetitions for #148

### DIFF
--- a/chatterbot/utils/queues.py
+++ b/chatterbot/utils/queues.py
@@ -1,0 +1,30 @@
+class ResponseQueue(object):
+    """
+    This is a data structure like a queue.
+    Only a fixed number of items can be added.
+    Once the maximum is reached, when a new item
+    is added the oldest item in the queue will
+    be removed.
+    """
+
+    def __init__(self, maxsize=10):
+        self.maxsize = maxsize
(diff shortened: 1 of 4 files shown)


chatterbot/utils/queues.py
```diff
@@ -0,0 +1,30 @@
+class ResponseQueue(object):
```
kevin-brown: Is there any reason why you aren't using the Python [`queue`](https://docs.python.org/3/library/queue.html) here?

gunthercox: The name if this class may be misleading. It's more of a list. It needs to possible to check if an object is in the list, and it needs to have a size limit in such a way that the oldest object is removed when adding a new object. I'm not sure if overriding queue has any advantages in this case.

PR #175, opened 2016-05-20
tiny change

ozylol:


gunthercox:
I actually had the python syntax highlighting in there a while back. The issue is that it is valid markdown syntax on GitHub, but it is not a part of the markdown specification. I use pandoc in the projects `setup.py` file to convert markdown to restructured text format when I deploy to pypi. Unfortunately if I include the GitHub style syntax highlighting the page fails to render properly in pypi (https://pypi.python.org/pypi/ChatterBot).

ozylol:
Oh I see. Ok, thats fine then.

PR #201, opened 2016-06-25
Spanish translation

SirIvanMoReau:
Added Spanish translation and links to different translations in the
readme files. <3


SirIvanMoReau:
A problem... DX

PR #216, opened 2016-07-22
Added UTF-8 support in corpus

mbarisa:
Diacritics did not work with Croatian corpus and I think it should help with other languages as well. I am planning to create a pull request for Croatian corpus as well these days.

### DIFF
--- a/chatterbot/corpus/corpus.py
+++ b/chatterbot/corpus/corpus.py
@@ -28,8 +28,9 @@ class Corpus(object):
         Read and return the data from a corpus json file.
         """
         import json
+        import io
 
-        with open(file_name, encoding='utf-8') as data_file:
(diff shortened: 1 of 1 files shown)


chatterbot/corpus/corpus.py
```diff
@@ -29,7 +29,7 @@ def read_corpus(self, file_name):
         """
         import json
 
-        with open(file_name) as data_file:
+        with open(file_name, encoding='utf-8') as data_file:
```
kevin-brown: You probably want to use [`io.open`](https://docs.python.org/2/library/io.html#io.open) for Python 2 support.

mbarisa: Ok I used io.open now, now CI checks passed. 

Thanks

gunthercox:
Looks good :+1:

PR #225, opened 2016-08-09
Make the example compatible with python3

wwj718:


gunthercox:
Thank you :+1:

wwj718:
@gunthercox  I like your conversational dialog engine very much ， it is beautiful

PR #230, opened 2016-08-13
Add filters step to response selection process

gunthercox:
Filters are a new concept that make it more efficient to reduce the number of responses being processed. The filter methods are called before the logic adapter retrieves a set of possible response statements from the database. The significance is that filters operate using set logic so they will be faster than iterative logic that has to be used in many of the logic adapters. The goal of the filters is to filter out any statements that can't possible be valid responses before the results are passed to the logic adapters.

- [x] Tests
- [x] Documentation


gunthercox:
@navyad That is correct.

PR #250, opened 2016-08-24
Adding Flask Example Link to English ReadMe

chamkank:
I have created a [simple Flask app using ChatterBot](https://github.com/chamkank/flask-chatterbot) that would be a great starter template for anyone who uses Flask instead of Django. So I thought it might be helpful to leave a link to it alongside the Django example in the readme.

I will also add 2 more patches for the other readme files (in different languages) if this patch is accepted.


gunthercox:
Hi @chamkank, I would be happy to add the link.

PR #255, opened 2016-08-25
Avoid problems in str.format() with char accents

davizucon:
Hi, 

I down know if is new, but str.format(input_statement.text), have problems with accents chars.
I think the problem occurs only in TerminalAdapter,

Thanks,
Davi


gunthercox:
Hi @davizucon, I can't confirm what this fixes. Could you provide me with the steps needed to replicate the issue so I can test this?

davizucon:
Hi @gunthercox , If you input a char like "é" or "ã" with terminalAdapter, in log method it uses str.format(), if no unicode .format(() raise a error.

To reproduce, just clone master, and run with IN/OUT using TerminalAdapter.

Thanks,
Davi

davizucon:
@gunthercox , I`ll submit this fix again, there is a lot code changes here, I`ll clean up.

Issue #262, opened 2016-08-27
input statement vs existing statement


gsingers:
Not sure if there is a bug or intended, but in chatterbot.py, the following code implies that any extra_data that comes in on a statement that is _also_ on an existing statement loses the extra_data of the input statement.  In my case, I have an input adapter for Slack and I'm adding things like the user id and the channel the question came from, but I also have some training data and the extra info is being dropped.  

https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/chatterbot.py#L136

I think the extra_data should be propagated to the statement.

gunthercox:
I believe this issue has been fully resolved now. Feel fee to reopen it if I missed something.

PR #263, opened 2016-08-28
#262 fix: Pass through existing extra data from the input

gsingers:
See #262 for details.  Small patch to pass through extra_data from the original input statement.  This is useful for knowing where/how to respond, who asked the original question, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,7 @@ settings.py
 
 # Sphinx docs
 docs/_build/
+
+# IntelliJ
+.idea
+*.iml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -12,3 +12,7 @@ settings.py
 
 # Sphinx docs
 docs/_build/
+
+# IntelliJ
+.idea
```
kevin-brown: You probably want to also add these to your [global `.gitignore` file](https://davidwalsh.name/global-gitignore), so this doesn't have to be done for every project you contribute to.

gunthercox:
Thank you :+1: Looks good.

.gitignore
gsingers: yeah, good point.

PR #268, opened 2016-08-30
Fix arguments (+input_statement, statement_list) for most_frequent_re…

davizucon:
…sponse , and setting to lower case fuzz.ratio, sometimes the is not close enough to the right answer (my point of view).


gunthercox:
Looks great! :+1:

PR #269, opened 2016-08-30
Avoid problems in str.format() with char accents

davizucon:
If you input a char like "é" or "ã" with terminalAdapter, in log method str.format(), if no unicode .format(() raise a error.

To reproduce, just clone master, and run with IN/OUT using TerminalAdapter.

Thanks

### DIFF
--- a/chatterbot/utils/read_input.py
+++ b/chatterbot/utils/read_input.py
@@ -10,4 +10,8 @@ def input_function():
         user_input = str(raw_input())
     else:
         user_input = input()
+
+    if user_input:
+        user_input = user_input.decode('utf-8')
+        
     return user_input


gunthercox:
Hi @davizucon, sorry about the delay in reviewing this. I was able to confirm the issue, but I'm going to have to make a small change to the pull request so that it doesn't break Python 3 compatibility.

Python 3 has an improved way of dealing with strings which eliminates issues with unicode characters, however it breaks backwards compatibility with the string decode method.

Python 3:
```python
'é'.decode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode'
```

Python 2.7:
```python
'é'.decode('utf-8')

>>> u'\xe9'
```

gunthercox:
This has been closed by https://github.com/gunthercox/ChatterBot/commit/35760762280d082f777545dc11ca2be0355ecfaf

PR #272, opened 2016-09-03
approximate sentence match logic adapter and its docs 

sumn2u:
Jaccard index is found to have better performance in  approximate sentence matching so implement it.

### DIFF
--- a/chatterbot/adapters/logic/approximate_sentence_match.py
+++ b/chatterbot/adapters/logic/approximate_sentence_match.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+# Imports
+from .base_match import BaseMatchAdapter
+import nltk.corpus
+import nltk.tokenize.punkt
+import nltk.stem.snowball
+from nltk.corpus import wordnet
+import string
+from chatterbot.adapters import Adapter
+
+class ApproximateSentenceMatchAdapter(BaseMatchAdapter):
+    """
+    The Jaccard index is composed of a numerator and denominator.
+    In the numerator, we count the number of items that are shared between the sets.
+    In the denominator, we count the total number of items across both sets.
+    Let’s say we define sentences to be equivalent if 50% or more of their tokens are equivalent.  Here are two sample sentences:
+        The young cat is hungry.
+        The cat is very hungry.
+    When we parse these sentences to remove stopwords, we end up with the following two sets:
+        {young, cat, hungry}
+        {cat, very, hungry}
+    In our example above, our intersection is {cat, hungry}, which has count of two.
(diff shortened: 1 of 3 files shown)


chatterbot/adapters/logic/approximate_sentence_match.py
```diff
@@ -0,0 +1,87 @@
...
+                # Use a randomly picked statement
+                return 0, self.context.storage.get_random()
+            else:
+                raise self.EmptyDatasetException()
+
+        confidence = -1
+        sentence_match = input_statement
+        # Find the  matching known statement
+        for statement in statement_list:
+            ratio = self.is_ci_lemma_stopword_set_match(input_statement.text, statement.text)
```
gunthercox: I'm not sure if this works. Where is `ratio` being used?

sumn2u:
ratio is added here if there is ratio it will return result :)

chatterbot/adapters/logic/approximate_sentence_match.py
```diff
@@ -0,0 +1,90 @@
...
+            return (pos_tag[0], wordnet.NOUN)
+        elif pos_tag[1].startswith('R'):
+            return (pos_tag[0], wordnet.ADV)
+        else:
+            return (pos_tag[0], wordnet.NOUN)
+
+
+    def is_ci_lemma_stopword_set_match(self,a, b, threshold=0.5):
+        """Check if a and b are matches."""
+        print("ask",a)
```
gunthercox: This print statement will have to be removed before this gets merged.

chatterbot/adapters/logic/approximate_sentence_match.py
```diff
@@ -0,0 +1,90 @@
...
+        confidence = -1
+        sentence_match = input_statement
+        # Find the  matching known statement
+        for statement in statement_list:
+            ratio = self.is_ci_lemma_stopword_set_match(input_statement.text, statement.text)
+            if ratio:
+                closest_match = statement
+            else:
+                closest_match = statement
+        return 50, closest_match
```
gunthercox: Should this be returning `ratio` instead of a fixed value of 50?

sumn2u: Ratio is a Boolean value  since our threshold value is 0.5 which is 50 we make 50 :)

gunthercox: 0.5 would actually be correct here. It appears I never documented it anywhere but confidence values are always between 0 and 1.

sumn2u: Yup so I put 0.5 over there :)

gunthercox:
Thank you! Looks good :smile:

PR #275, opened 2016-09-04
base filter is abstract

navyad:
Filter is made abstract so the it called be instantiated.


gunthercox:
Hi @navyad, I'm not sure if there are any significant advantages by making the base Filter class abstract. I appreciate the suggestion but I'm going to close this ticket off for now. Feel free to message me if I missed something significant regarding this.

PR #293, opened 2016-09-14
add `__str__` method to models

hobson:
Makes admin interface much more intuitive to work with.

### DIFF
--- a/chatterbot/ext/django_chatterbot/models.py
+++ b/chatterbot/ext/django_chatterbot/models.py
@@ -8,7 +8,6 @@ class Statement(models.Model):
         unique=True,
         blank=False,
         null=False,
-        default='<empty>',
         max_length=255
     )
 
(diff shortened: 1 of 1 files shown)


chatterbot/ext/django_chatterbot/models.py
```diff
@@ -2,16 +2,34 @@
 
 
 class Statement(models.Model):
+    """A short (<255) chat message, tweet, forum post, etc"""
 
     text = models.CharField(
         unique=True,
         blank=False,
         null=False,
+        default='<empty>',
```
gunthercox: I'm not sure about having a default response specified because it will alter how the chat bot handles knowledge it has accumulated for conversations. If a default of `'<empty>'` is set, the chat bot will then be able to respond to a user's input with a response of '<empty>'.

Was there a reason for adding this?

gunthercox:
Hi @hobson, these changes look great. I just had question about the changes, once that is resolved I think this should be good to merge.

chatterbot/ext/django_chatterbot/models.py
hobson: Good point. Hadn't thought of that. I was just trying to get the amin interface to display something on the Statements list view so that the empty statements have a link that can be clicked to "walk the graph" of statement response. Will remove it.

PR #300, opened 2016-09-18
Improve the Bayes classifier in time_adapter.py

rileyjshaw:
It looks like `"do you know the time"` was accidentally added to both the "hit" and "miss" lists. This commit ensures it's classified as a match.


gunthercox:
Thank you! Looks good 👍

PR #317, opened 2016-10-01
Use ThreadPoolExecutor for processing multiple logic adapters concurrently

### LINKED ISSUE #110: Multithreading support during logic adapter selection
To increase the speed of processing, multithreading support is really needed.

In the future, when more logic adapters are added including more advanced ones (such as a developer assistant), https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/adapters/logic/multi_adapter.py#L11 will consume a lot of computing resources. If this is all in one thread, it could cause a slowdown of the program this is very noticeable.

Here is my solution: We enable threading at that stage of the processing. In addition, if that is not enough, individual logic adapters can also implement threading, but that is for another issue

gunthercox:
Hypothetically closes #110

### DIFF
--- a/chatterbot/adapters/logic/multi_adapter.py
+++ b/chatterbot/adapters/logic/multi_adapter.py
@@ -1,4 +1,5 @@
 from .logic_adapter import LogicAdapter
+from concurrent.futures import ThreadPoolExecutor
 
 
 class MultiLogicAdapter(LogicAdapter):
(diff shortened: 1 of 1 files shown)


chatterbot/adapters/logic/multi_adapter.py
```diff
@@ -24,9 +48,11 @@ def process(self, statement):
...
         max_confidence = -1
 
-        for adapter in self.adapters:
-            if adapter.can_process(statement):
-                confidence, output = adapter.process(statement)
+        usable_adapters = self.get_usable_adapters(statement)
+
+        # with ProcessPoolExecutor(max_executors) as executor:
+        with ThreadPoolExecutor(max_workers=len(usable_adapters)) as executor:
+            for confidence, output, adapter in executor.map(self.execute_adapter_process, usable_adapters):
```
kevin-brown: I believe since this is happening sequentially (locked by the for loop), this isn't actually giving you any extra performance. Unless `.map` doesn't guarantee ordering, in which case this makes sense.

gunthercox:
I believe the reason that Django-related tests are currently failing for these changes because this is causing multiple database transactions to occur in a single request. https://groups.google.com/forum/#!msg/django-users/5TRPydzRiJg/u7Q18h8ccxsJ

chatterbot/adapters/logic/multi_adapter.py
gunthercox: The executor's `map` function does execute the function asynchronously with each of the values in the iterable. (https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map). From what I understand, the function execution via `map` is asynchronous, but the for loop is still synchronous.

PR #322, opened 2016-10-04
added hindi corpus

HarshGrandeur:
Added corpus of HINDI language, will be adding more of conversations in hindi language


gunthercox:
Thank you, looks good :+1:

PR #332, opened 2016-10-09
Pass logic adapter confidence to the output adapter

gunthercox:
This makes it possible for the output adapter to make a decision about if some kind of action (such as an API call) should be taken based on that confidence value.


rmdort:
Would this will help in responding with `I dont understand your question` for statements that don't match.

gunthercox:
@rmdort Sure, it comes down to implementing that in your own output adapter, but that could definitely be done.

PR #333, opened 2016-10-09
Fix init class name

rmdort:
Incorrect class name

````
super(ClosestMatchAdapter, self).__init__(**kwargs)
NameError: global name 'ClosestMatchAdapter' is not defined
````


gunthercox:
Looks good 👍

PR #334, opened 2016-10-09
fix readme es adding a dot at the end

sergiosalfer:


gunthercox:
I don't believe this is an error.

PR #337, opened 2016-10-10
Force training even if read_only is True

rmdort:
This is just a proposal. 

Scenario
Most of my clients want a training corpora but they dont want bots to auto-learn.

So when i am starting up the bot

````
bot = ChatBot('name', read_only=True)
bot.train('english') # Add corpus even if read_only is True
````

What do you think?

### DIFF
--- a/chatterbot/adapters/storage/mongodb.py
+++ b/chatterbot/adapters/storage/mongodb.py
@@ -173,9 +173,9 @@ class MongoDatabaseAdapter(StorageAdapter):
 
         return results
 
-    def update(self, statement, force=False):
+    def update(self, statement, **kwargs):
         from pymongo import UpdateOne, ReplaceOne
-
+        force = kwargs.get('force', False)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: chatterbot/adapters/storage/mongodb.py [hunk 1]
```diff
@@ -173,11 +173,11 @@ def filter(self, **kwargs):
 
         return results
 
-    def update(self, statement):
+    def update(self, statement, force = False):
```
karanluthra: Is this style consistent with rest of the code base?
I have seen foo=bar in method parameters at quite a few places in this code base. 
Maybe @gunthercox could take the call.

gunthercox: I agree with @karanluthra, there should be no spaces when doing assignments in method parameters.

gunthercox:
Tests are currently failing on this pull request so unfortunately I cannot merge it yet.

PR #362, opened 2016-10-26
Closest subtree matching

### LINKED ISSUE #276: Tie breaking method based on the current conversation history
ChatterBot currently has a [few tie breaking methods](https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/adapters/logic/mixins.py) that can be used to determine how it selects a response when there are multiple equivalent possible responses.

Hypothetically, the current conversation that the user is having with the chat bot could be compared to past conversations for similarity. This information could then be used to select a response from the list of possible options.

gunthercox:
This adds methods that make it possible to find the closest match of a sequence in a tree that represents different possible configurations of that sequence.

- [x] Conversation objects needs to be implemented

Closes #276


vkosuri:
Master, any updates about this pr

gunthercox:
@vkosuri Not currently. Before anything in this pull request can be made to work, ChatterBot's current concept of "Sessions" need to be replaced by database-backed "Conversations".

vkosuri:
@gunthercox Master, could you please guide me how to do this replacement?

vkosuri:
How to do this task

> "Sessions" need to be replaced by database-backed "Conversations".

gunthercox:
@vkosuri I will write up a guide as soon as I can. It is going to be a somewhat long process to make the switch because there is so much involved.

Issue #350, opened 2016-10-16
ClosestMeaningAdapter nltk_data download loop


Nixellion:
Whenever I try to use ClosestMeaningAdapter, it goes into download loop like this:

```
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package punkt is already up-to-date!
[nltk_data] Downloading package stopwords to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package stopwords is already up-to-date!
[nltk_data] Downloading package wordnet to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package wordnet is already up-to-date!
[nltk_data] Downloading package punkt to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package punkt is already up-to-date!
[nltk_data] Downloading package stopwords to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package stopwords is already up-to-date!
[nltk_data] Downloading package wordnet to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package wordnet is already up-to-date!
[nltk_data] Downloading package punkt to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package punkt is already up-to-date!
[nltk_data] Downloading package stopwords to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package stopwords is already up-to-date!
[nltk_data] Downloading package wordnet to
[nltk_data]     C:\Users\nixes\AppData\Roaming\nltk_data...
[nltk_data]   Package wordnet is already up-to-date!
```

However nltk is downloaded. Is it me, or is there a bug?

gunthercox:
I've seen this before, but it appears to be somewhat sporadic. Occasionally data files will be downloaded for NLTK even when they have already been downloaded. I will look into the issue.

Nixellion:
Thanks. Happens all the time here :(

vkosuri:
I believe you were seeing this looping message due to https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/utils/wordnet.py#L18

Usually we don't have privileges default nltk download directories, obviously We will get LookupError Exception, 

[...]

PR #369, opened 2016-10-30
nltk_data download loop

vkosuri:
https://github.com/gunthercox/ChatterBot/issues/350


gunthercox:
Looks great! Thank you.

PR #370, opened 2016-10-30
Master

ThomasKoscheck:
Added german translating of corpus.math_word and corpus.greetings


gunthercox:
Looks great, thank you.

PR #375, opened 2016-10-31
Add check for logic adapter agreement

### LINKED ISSUE #169: Check for subadapter agreement in MultiLogicAdapter
ChatterBot's MultiLogicAdapter module makes it possible to get a response based the output of any number of logic adapters. Currently, ChatterBot returns the response with the highest confidence value. 

One concept I am considering is the idea that if multiple adapters agree on the same statement, then that statement is more likely to be the correct response.

There are a few ways this could be implemented.
1. The first is to simply add a check to see if any adapters have returned the same response.
2. The second requires a greater number of changes to implement. This would involve redesigning logic adapters as a pipeline process. This means that each adapter would be in charge of removing a number of responses from a set of possible correct responses before passing that set to the next logic adapter. One advantage is that this would reduce some repetitive queries that are currently made in multiple logic adapters. One disadvantage is that the pipeline is not a process that could be made to run in parallel in the future.
### Pipeline notes
- If a logic adapter returns 0 results, the results from the previous adapter should be passed on instead. This is a measure to make sure that meaningful data is being processed since the last step in the pipeline would likely be to return a random response if no responses could be determined.
- Some logic adapters that generate dynamic results such as the math and time adapters may need to insert statements into the data they return. They could be skipped if their `can_process` methods returned False.

gunthercox:
Closes #169

### DIFF
--- a/chatterbot/adapters/logic/multi_adapter.py
+++ b/chatterbot/adapters/logic/multi_adapter.py
@@ -1,4 +1,5 @@
 from .logic_adapter import LogicAdapter
+from collections import Counter
 
 
 class MultiLogicAdapter(LogicAdapter):
@@ -21,12 +22,14 @@ class MultiLogicAdapter(LogicAdapter):
 
         :param statement: The input statement to be processed.
         """
+        results = []
         result = None
         max_confidence = -1
 
         for adapter in self.adapters:
             if adapter.can_process(statement):
                 confidence, output = adapter.process(statement)
+                results.append((confidence, output, ))
 
                 self.logger.info(
                     u'{} selected "{}" as a response with a confidence of {}'.format(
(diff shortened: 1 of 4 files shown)


rmdort:
@gunthercox Just wanted to know more about this? Could you give an example when this will occur?

gunthercox:
Sure, consider a scenario where multiple logic adapters are being used. In a case where the outputs generated by the logic adapters are as follows:

|    Confidence    |    Statement   |
|------------------|----------------|
|  0.2  |  'Good morning.'  |
|  0.5  |  'Good morning.'  |
|  0.7  |  'Good night.'   |

In this case, two of the logic adapters have generated the same result. The idea behind these changes is to treat the fact that the two adapters have agreed on a response as a significant indicator that this particular response has a greater probability of being a more accurate response to the input.

PR #380, opened 2016-11-04
Master

ThomasKoscheck:
Added uft-8 coding for using german umlauts


chatterbot/corpus/data/german/conversations.corpus.json
```diff
@@ -1,84 +1,87 @@
+# -*- coding: utf-8 -*-
```
gunthercox: This is a [Python unicode header](https://www.python.org/dev/peps/pep-0263/), usually they go at the top of Python files. It looks like you've added it to a `json` file. This isn't valid json and I'm not ever sure that ChaterBot's json loader will read it correctly.

gunthercox:
@ThomasKoscheck I'm going to close this pull request due to the aforementioned issues.

PR #424, opened 2016-11-27
Add session management

### LINKED ISSUE #372: Session management
@gunthercox any plans adding sessions management to ChatterBot?

gunthercox:
- [x] Create new session objects
- [x] Allow integration with Django sessions
- [x] Add tests for sessions
- [x] Update documentation

Closes #372

### DIFF
--- a/chatterbot/chatterbot.py
+++ b/chatterbot/chatterbot.py
@@ -191,7 +191,7 @@ class ChatBot(object):
         :rtype: Statement
         """
         if not session_id:
-            session_id = self.default_session.uuid
+            session_id = str(self.default_session.uuid)
 
         input_statement = self.input.process_input_statement(input_item)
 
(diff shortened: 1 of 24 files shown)


vkosuri:
@gunthercox will it consider other bot session objects, for example gitter/microsoft session object was expired then chatterbot session object will expire?, will this case is considered?

gunthercox:
If two bots are created, each will have sessions separate from each other.

```
bot1 = ChatBot()
bot2 = ChatBot()
```

For each bot, the sessions _could_ be shared across multiple adapters as long as there is a shared session id. Right now this is something that the developer would have to handle manually because a `ChatBot` instance has no way of knowing that a Gitter user is the same person as a user on the Microsoft Bot Framework. To handle this, the developer would likely have to store associations between a unique identifier and a user's verified usernames for each service (Gitter, HipChat, Microsoft Bot Framework, etc.).

Hopefully this answers your question. I'm planning to add more documentation on this to hopefully make it clearer.

vkosuri:
Nice, Thank you

PR #449, opened 2016-12-01
New math operations

### LINKED ISSUE #320: Additional operations for the Math logic adapter
- square root
- logarithms
- Pi

le-corentin:
The changes are:

- added support for **mathematical constants and** some numpy **functions** (closes #320)
- some **refactoring** of the chunk processing (nested `if`/`else` statements)
- **separating words** using a operators rather than spaces (allowing strings such as `"log(3+4)"` to be processed, instead of just `"log ( 3 + 4 ) "`)

I hope this is in keeping with coding standards of the project, especially the addition of a class attribute (`functions` in `MathematicalEvaluation`). Also, there could be more [math functions](https://docs.scipy.org/doc/numpy/reference/routines.math.html) and [constants](https://en.wikipedia.org/wiki/Mathematical_constant#Table_of_selected_mathematical_constants).

### DIFF
--- a/chatterbot/logic/mathematical_evaluation.py
+++ b/chatterbot/logic/mathematical_evaluation.py
@@ -5,6 +5,7 @@ import re
 import os
 import json
 import decimal
+import numpy
 
 
 class MathematicalEvaluation(LogicAdapter):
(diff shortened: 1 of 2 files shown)


le-corentin:
Thanks a lot for your enthusiasm!

Issue #308, opened 2016-09-26
Django App statement -> response relationship is reversed


hobson:
To reproduce:
1. `runserver` on the example django app.
2. visit localhost:8000/admin to log in and view the Response model
3. click on the chatterbot app and the `Responses` table
4. click on one of the statements
5. bug: the field labeled `statement` is what you typed in response to the text shown under the `response`

The Response table should hold the foreign keys to the statement and the response that followed it but the field names are reversed so that Response.statement holds the response and Response.response holds the statement that was being responded to.

Propose renaming `models.Response.response` to `models.Response.response_to` as an interim fix. Need to double check that the logic adapter is being sent the right response_to text in order to formulate it's responses. Here's my attempt at fixing it and adding better visability into the data with the admin interface: https://github.com/totalgood/ChatterBot/blob/dev/chatterbot/ext/django_chatterbot/models.py

Unfortunately I just discovered this confusing name "bug" after developing a lot of additional new features, and I won't be able to break it up into manageable PRs for at least 2 weeks. Demoing it at PyDX next weekend.

PR #471, opened 2016-12-07
Tighter Django integration

### LINKED ISSUE #464: Do the Django Statement and Response models need to be converted back into ChatterBot Statement and Response objects?
Right now, when using ChatterBot with Django, ChatterBot's Django storage adapter converts the Django Statement and Response model objects back into ChatterBot Statements and Response objects.

Because both objects share similar structures, the conversion might not actually be necessary and ChatterBot can use the Django objects themselves. This change would greatly improve performance when processing large amounts of data.

gunthercox:
This makes changes so that if ChatterBot is being used with Django, the Django Statement and Response models can be used in place of ChatterBot's Statement and Response objects. Previously the storage adapter converted incoming statement objects into Django model objects before saving and/or returning them. Now this conversion is no longer required.

This is going to need further testing before it gets merged.

Items that need to be checked off before this can be merged:

- [x] Make sure ChatterBot statements/responses and the Django models have the same methods
- [x] Make sure logic adapters all work with in Django
- [x] Make sure input adapters all work in Django
- [x] Make sure output adapters all work in Django

Closes #464


vkosuri:
@gunthercox Any plan to fix this issue https://github.com/gunthercox/ChatterBot/issues/308

gunthercox:
@vkosuri I'm planning on taking care of #308 in a pull request after this one gets merged. I'm pretty sure the "reversal" is just an issue with confusing variable names.

Issue #494, opened 2016-12-12
wsgi issues


vkosuri:
I have found some issue will submit PR.

gunthercox:
Closed by #505

PR #495, opened 2016-12-12
Fix wsgi issues in django app

vkosuri:
This fix will address https://github.com/gunthercox/ChatterBot/issues/494 and fix fallowing issues also.

1. Adding glyphicons halflings fonts will solve while using wsgi 
2. Adding whitenoise and gunicorn for wsgi
3. Adding whitenoise settings to export static files while using wsgi


.travis.yml
```diff
@@ -18,6 +18,7 @@ install:
   - pip install .
   - pip install -r test-requirements.txt
   - pip install -r requirements.txt
+  - pip install -r examples/django_app/requirements.txt
```
gunthercox: This actually shouldn't be installed during testing. The first line (`pip install .`) installs the ChatterBot codebase for the current commit being tested. `pip install -r examples/django_app/requirements.txt` will install the current version of ChatterBot from PyPi.

Additionally, the tests also test on different versions of Django to ensure compatibility for each supported version. Installing from the example app requirements would alter the version of Django that gets installed to the test environment on Travis CI.

examples/django_app/example_app/static/fonts/glyphicons-halflings-regular.svg
```diff
@@ -0,0 +1,288 @@
+<?xml version="1.0" standalone="no"?>
```
gunthercox: Since the Bootstrap Glyphicons aren't actually used in the example app, I'd rather not add them. I realize that a message about them not being found is currently displayed in the networking output. It might be better to just remove the line in the stylesheets that loads the Glyphicons.

examples/django_app/example_app/wsgi.py
```diff
@@ -10,7 +10,9 @@
 import os
 
 from django.core.wsgi import get_wsgi_application
+from whitenoise.django import DjangoWhiteNoise
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_chatterbot.settings")
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "example_app.settings")
```
gunthercox: This is a good change and should get pulled in.

examples/django_app/requirements.txt
```diff
@@ -1,2 +1,4 @@
 django>=1.8.0,<2.0.0
 chatterbot>=0.5.0
+gunicorn
+whitenoise
```
gunthercox: Was there a particular reason gunicorn and whitenoise are needed in the _example_ app?

.travis.yml
vkosuri: Agreed i'll remove it.

examples/django_app/requirements.txt
vkosuri: If the example_app is going to use web service/hosting definitely. If web serive it won't consider static files, If we want serve static files then we have to use ``whitenoise`` module.

@gunthercox There is other way, I'll remove the changes from here and add these steps in ``django.rst`` documentation page.

gunthercox: I think adding these steps to the Django documentation sounds like a good solution, it allow the example app to remain simple while still providing recommendations to developers who are looking to create a hosted solution on a server.

Issue #497, opened 2016-12-12
django chat response aren't converting into json objects


vkosuri:
Captured Traceback of thsi errror

``` Bash
Traceback:  

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/exception.py" in inner
  39.             response = get_response(request)

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py" in _legacy_get_response
  249.             response = self._get_response(request)

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/app/.heroku/python/lib/python2.7/site-packages/django/views/generic/base.py" in view
  68.             return self.dispatch(request, *args, **kwargs)

File "/app/.heroku/python/lib/python2.7/site-packages/django/views/generic/base.py" in dispatch
  88.         return handler(request, *args, **kwargs)

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/ext/django_chatterbot/views.py" in post
  67.         response_data = self.chatterbot.get_response(input_data, chat_session.id_string)

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/chatterbot.py" in get_response
  98.         statement, response, confidence = self.generate_response(input_statement, session_id)

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/chatterbot.py" in generate_response
  122.         confidence, response = self.logic.process(input_statement)

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/logic/multi_adapter.py" in process
  37.                 confidence, output = adapter.process(statement)

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/logic/best_match.py" in process
  59.         confidence, closest_match = self.get(input_statement)

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/logic/best_match.py" in get
  23.         statement_list = self.chatbot.storage.get_response_statements()

File "/app/.heroku/python/lib/python2.7/site-packages/chatterbot/storage/storage_adapter.py" in get_response_statements
  88.         statement_list = self.filter()

[...]

PR #498, opened 2016-12-12
response into JSON object

vkosuri:
Somehow the current djanog response isn't converting into JSON object, it always returns. https://github.com/gunthercox/ChatterBot/issues/497 issue will fix this error.

### DIFF
--- a/chatterbot/ext/django_chatterbot/views.py
+++ b/chatterbot/ext/django_chatterbot/views.py
@@ -66,6 +66,13 @@ class ChatterBotView(ChatterBotViewMixin, View):
 
         response_data = self.chatterbot.get_response(input_data, chat_session.id_string)
 
+        try:
+            unicode = unicode
+        except NameError:
+            # Must be python 3
+            unicode = str
+            basestring = (str, bytes)
+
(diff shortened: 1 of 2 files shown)


vkosuri:
@gunthercox sure will work on those two tests, One question, where do i need to add these test either in django example app [tests ](https://github.com/gunthercox/ChatterBot/blob/master/examples/django_app/tests/test_example.py)or in [django_tests](https://github.com/gunthercox/ChatterBot/blob/master/tests_django/test_views.py)

gunthercox:
These tests should be in the `django_tests` directory. You can run these django-specific tests with:

```
python runtests.py
```

vkosuri:
@gunthercox i am little busy with other tasks, It may take another day or two, Could you please let me are there any ETAs for new release?

gunthercox:
I don't have an ETA yet, I'm possibly looking to do one either this weekend or next week.

vkosuri:
appologies for many failures, needs to learn  ``python3`` 📦

gunthercox:
This looks great, thank you for adding the test cases. They help make sure that this functionality never breaks again in the future.

PR #500, opened 2016-12-13
Deprecate OutputFormatAdapter

gunthercox:
This removes the need to replicate basic methods and properties that are available on the statement object. The `OutputFormatAdapter` will eventually be removed.


chatterbot/ext/django_chatterbot/settings.py
```diff
@@ -10,8 +10,7 @@
     'name': 'ChatterBot',
     'storage_adapter': 'chatterbot.storage.DjangoStorageAdapter',
     'input_adapter': 'chatterbot.input.VariableInputTypeAdapter',
-    'output_adapter': 'chatterbot.output.OutputFormatAdapter',
-    'output_format': 'json'
```
DynasticFX: how to set it now to text?

gunthercox:
*@DynasticFX*  You can get the text using the `.text` attribute on the
response that is returned.

PR #507, opened 2016-12-15
Fix django view json loads error

barseghyanartur:
Discovered a bug when testing the [django_chatterbot](https://github.com/gunthercox/django_chatterbot). This patch solves the problem.

### DIFF
--- a/chatterbot/ext/django_chatterbot/views.py
+++ b/chatterbot/ext/django_chatterbot/views.py
@@ -1,4 +1,6 @@
 import json
+from urlparse import parse_qs
+
 from django.views.generic import View
 from django.http import JsonResponse
 from chatterbot import ChatBot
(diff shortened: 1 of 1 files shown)


gunthercox:
Hi @barseghyanartur Thank you for the pull request. The example at https://github.com/gunthercox/django_chatterbot is no longer actively maintained. A while back the example was moved to: https://github.com/gunthercox/ChatterBot/tree/master/examples/django_app

It is very likely that the issue you encountered no longer exists in the updated version of the app. Would you mind confirming that?

barseghyanartur:
@gunthercox:

I see. I'll confirm it tomorrow and let you know. Thanks!

Perhaps a notice "This is no longer actively maintained" with link to the well maintained version would be useful then.

PR #513, opened 2016-12-17
Allow chat bot to track individual conversations

### LINKED ISSUE #366: Multiple knowledge themes
From what I understand, right now all the conversations of the chatbot are added to one database.

But what about 'injecting' and 'removing' certain knowledge?
For example, a chatbot is trained from live conversations with users, but at some point I want to inject into him dialogues from a certain book, to allow him to use answers from that book along with those trained from actual people.

But then at some point I may want to make him forget only that book's dialogues, not the whole db.

Or maybe I want him to talk to one user using only answers from book N, but to another user using answers from book X, but also using conversation training Y for both of them.

Is it currently possible? If not, it may be an interesting feature to think about.

gunthercox:
- Required for #362 to search for similar conversations.
- Closes #366 by adding `Conversation` objects.

**Tasks remaining:**
- [ ] Create a migration plan to move to the new configuration
- [ ] Update documentation
- [ ] Create and update tests


chatterbot/conversation/conversation.py
```diff
@@ -0,0 +1,20 @@
+import uuid
+from chatterbot.queues import ResponseQueue
```
gunthercox: `ResponseQueue` will need to be removed and the methods are going to be updated to use database queries.

PR #548, opened 2017-01-02
nltk.data.find argument needs to be a path, not just a filename

Gaboose:
`nltk_download_corpus` always downloads nltk resources even if it's already there.

It's because neither `nltk.data.find('stopwords.zip')` nor `nltk.data.find('stopwords')` finds the resource. Maybe `nltk_data` directory tree was flat at some point in the past, but right now it gets populated like this: 

```bash
- corpora
  |- stopwords
  |- stopwords.zip
  |- wordnet
  |- wordnet.zip
- sentiment
  |- vader_lexicon.zip
- tokenizers
  |- punkt
  |- punkt.zip
```

and so only a call like `nltk.data.find('corpora/stopwords')` works.

### DIFF
--- a/chatterbot/chatterbot.py
+++ b/chatterbot/chatterbot.py
@@ -77,10 +77,10 @@ class ChatBot(object):
         from .utils import nltk_download_corpus
 
         # Download required NLTK corpora if they have not already been downloaded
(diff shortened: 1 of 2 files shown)


gunthercox:
Thank you 👍

PR #549, opened 2017-01-02
Allow custom file paths to ChatterBot corpus files

### LINKED ISSUE #467: Custom paths for corpus data
I think it would be good option to specify custom paths for corpus data, currently its system specified path only

```Python
class Corpus(object):

    def __init__(self):
        current_directory = os.path.dirname(__file__)
        self.data_directory = os.path.join(current_directory, 'data')
```

gunthercox:
Closes #467


vkosuri:
👍

PR #569, opened 2017-01-11
Convert clean_whitespace utility into a preprocessor

gunthercox:
This adds a new feature called _preprocessors_. These are simple functions that modify the input statement before it is processed by the logic adaper.


vkosuri:
👍

PR #516, opened 2016-12-18
AIML english corpus updation into chatterbot

vkosuri:
By looking into Repo https://github.com/drwallace/aiml-en-us-foundation-alice, i think we need update many conversations. I am started updating few important one

- [x] science
- [x] food
- [x] drugs
- [x] bot profile
- [x] psychology
- [x] politics
- [x] humor
- [x] history
- [x] gossip
- [x] emotion
- [x] ai
- [ ] knowledge
- [x] literature
- [x] movies
- [x] money
- [x] sports

PR #585, opened 2017-01-15
Some minor text changes

martmists:
Update for #516, fixes everything pointed out by @gunthercox


martmists:
Tests on this branch seem outdated, so tests will fail:
`AssertionError: '01-01-2017' != '01-01-2016'`

PR #595, opened 2017-01-17
Store sessions in database

gunthercox:
This pull request will make changes to store sessions in the database. Sessions will eventually be used to track individual conversations. The session needs to be stored so that statements can reference the individual conversation that they are a part of.

For #513


meanchun:
i think Store sessions in database is  important. why it not be used?

gunthercox:
@meanchun You are correct, it is important. Additional work is still required to add this functionality.

PR #596, opened 2017-01-17
default db file

vkosuri:
I think it is good to specify default file name of ListTrainer

### DIFF
--- a/docs/training.rst
+++ b/docs/training.rst
@@ -43,7 +43,8 @@ Training via list data
 For the training, process, you will need to pass in a list of statements where the order of each statement is based
 on it's placement in a given conversation.
 
-For example, if you were to run bot of the following training calls, then the resulting chatterbot would respond to
+For example, if you were to run bot of the following training calls, then the resulting chatterbot would creates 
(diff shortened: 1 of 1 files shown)


docs/training.rst
```diff
@@ -43,7 +43,8 @@ Training via list data
 For the training, process, you will need to pass in a list of statements where the order of each statement is based
 on it's placement in a given conversation.
 
-For example, if you were to run bot of the following training calls, then the resulting chatterbot would respond to
+For example, if you were to run bot of the following training calls, then the resulting chatterbot would creates 
+a default database in current directory with name ``database.db`` and cahtterbot will respond to
```
gunthercox: What if they have specified a different storage adapter? Do you think this is too specific? The training class doesn't just save to a `database.db` file, it saves to whatever storage endpoint is defined by the chat bot's storage adapter. So essentially, the `database.db` file has nothing to do with training, it all depends on the storage adapter settings.

vkosuri: How about this 

a default database will be created with the name ``database.db`` in current directory if haven't configure any database endpoint.

vkosuri: @gunthercox Master any updates?

PR #597, opened 2017-01-17
Add custom JSON adapter

### LINKED ISSUE #592: Add internal support for different JSON libraries.
As mentioned in #42, cjson is used when installed automatically to speed up. However, cjson does not yet support Python 3. This is why I'm proposing to add a similar method for a Py3 version of a fast JSON library, such as ujson or python-rapidjson.

martmists:
The purpose of this adapter is to make use of different JSON parsing libraries, such as ujson or python-rapidjson. 

Example usage:
```python
chatbot = chatterbot.ChatBot(
    "Example bot",
    trainer='chatterbot.trainers.ChatterBotCorpusTrainer',
    storage_adapter={
        "import_path": "chatterbot.storage.CustomJsonFileStorageAdapter",
        "dump_func":"ujson.dumps",
        "load_func":"ujson.loads"
    }
)
chatbot.train("chatterbot.corpus.english")
```

As you can see, all that has to be done is to make `storage_adapter` a dict with an `import_path`, `dump_func` and `load_func`.
both dump and load functions default back to the stdlib json module. Note that `dumps` and `loads` WILL NOT WORK.

This has successfully been tested, however I do not know if I messed anything up by accident which could make it function less like it should.

Fixes #592


martmists:
@gunthercox feel free to merge whenever

martmists:
Noted. The reason I redid most of the I/O handling is because the current JSON adapter uses I/O about 4 times for every `filter` call, causing it to slow down a lot. Once for every `update` call looked too much as well, so I set it up to save for every `get_response`. I will see if changing jsondb's source changes any of this, but I'm fairly sure it's the constant I/O that slows it down.

martmists:
I made a PR to jsondb [here](https://github.com/gunthercox/jsondb/pull/7). Using my PR as a test, I found that, despite it being a little faster due to it using ujson, is still relatively slow. My guess is that the amount of File I/O calls is too high, which is why I propose to either make the amount of jsondb calls lower, or implement the custom adapter which makes less I/O calls.

martmists:
Above two commits were made to support libraries such as `rapidjson`, which only supports `dumps` and `loads` calls.

vkosuri:
@martmists before pushing your commits try to runs tests your locally.

martmists:
@vkosuri if you could explain to me how I will, but as said before I have no clue how to test

vkosuri:
@martmists I think this will help you http://chatterbot.readthedocs.io/en/stable/testing.html

martmists:
I'll probably rewrite this storage adapter to use any dumping/loading method as to enable it to use for example the `erlpack` module.

gunthercox:
Hi @martmists, I apologize about the amount of time this pull request has remained open for. I'm going to close it off because I'm moving the project away from `jsondb` in favor of supporting popular, existing types of databases.

The JsonDatabaseAdapter in ChatterBot will be supported until the 0.7 version release but I'm not looking to add additional JSON type adapters anymore. I appreciate the work you put into this pull request and I'm sorry that I could not be merged into the project. If you still want to make it available for others to use you can always set it up as it's own Python package on PyPi.

martmists:
I see, thanks for showing interest though!
I'm going to continue using this adapter myself, and I'll leave my fork of the repo up and up to date for people who want to use it as well.

PR #618, opened 2017-02-05
Enhancement of the decimal percent of similarity in levenshtein_distance

addelll:


gunthercox:
Looks good. Thank you!

PR #633, opened 2017-02-24
Best match adapter to support other languages

cristobalsr:
Current synset comparer only works with english language.  I have made some changes to allow set the language using a best match logic adapter extension by two parameters lang and languages.


vkosuri:
@cristobalsr could you please add tests for this PR?

cristobalsr:
I'm going to close this pull request because the code is not completed yet.

PR #653, opened 2017-03-15
fixed split databases usage for django

web-maker:
Added boolean `use_django_models` option for ChatterBot settings to avoid problems with splitted databases usage.


gunthercox:
Looks good. Thank you @web-maker 👍

PR #685, opened 2017-04-12
Experimental support for a better JSON adapter.

Gustavo6046:
I know, it doesn't really work THAT well right now, but once one of us trace the origin we will fix it. The only bug? Most statements simply aren't saved. But then again I'm not the best code interpreter myself. :P


gunthercox:
Hi @Gustavo6046, it looks like there are a few issues with this storage adapter. I've also been thinking about the best way to solve the issue with this adapter being non-thread safe. The requirements necessary to create a proper JSON formatted database are enormous and if I started working on it I would end up with something close to Mongo DB (which is essentially a powerful JSON database).

At this point I've decided to steer the ChatterBot project away from using my original home-rolled file database and add support for SQLite which can still be used locally without any additional non-python dependencies and its developers have already solved the issues of read and write locking, querying for data, etc.

Gustavo6046:
That seems like a good idea, thanks.

PR #716, opened 2017-05-04
Keep a reference to a single session in SQL alchemy adapter

### LINKED ISSUE #707: SQLite database not being deleted after tests
I noticed that the test databases created by the SQLite storage adapter test cases are not being deleted when testing is concluded. As a result, several database files are created each time tests are run.

gunthercox:
This change makes it possible to close the session and delete the test database file when unit tests have finished running.

A few other minor changes have been made as well (see commit messages).

***

For testing, I've made changes so that a single database file is created before the set of test cases are run and then it is deleted after. The database is cleared for each test case. These changes should make it more efficient to run the tests. Closes #707

### DIFF
--- a/chatterbot/storage/sqlalchemy_storage.py
+++ b/chatterbot/storage/sqlalchemy_storage.py
@@ -103,7 +103,7 @@ class SQLAlchemyDatabaseAdapter(StorageAdapter):
         if not self.read_only and create:
             self.create()
 
-        Session = sessionmaker(bind=self.engine)
+        Session = sessionmaker(bind=self.engine, expire_on_commit=True)
         self.session = Session()
 
     def count(self):
--- a/chatterbot/storage/sqlalchemy_storage.py
+++ b/chatterbot/storage/sqlalchemy_storage.py
@@ -11,57 +11,56 @@ Base = None
 try:
     from sqlalchemy.ext.declarative import declarative_base
     Base = declarative_base()
+
+    class StatementTable(Base):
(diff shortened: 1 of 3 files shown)


chatterbot/storage/sqlalchemy_storage.py
```diff
@@ -98,30 +99,19 @@ def __init__(self, **kwargs):
...
+        create = self.kwargs.get("create", False)
 
-        if not self.read_only and self.drop_create:
-            _base.metadata.drop_all(self.engine)
-            _base.metadata.create_all(self.engine)
+        if not self.read_only and create:
+            self.create()
+
+        Session = sessionmaker(bind=self.engine)
+        self.session = Session()
```
gunthercox: @davizucon / @navyad I removed the `__get_session()` method and created a single instance of the session object. The reason I did this was because I think it was creating a new session each time `__get_session` was called (and not disposing of it).

I'm not very familiar with SQLAlchemy. If either of you have any insight on this or any of the other changes please let me know.

davizucon: No problem, I believe if you set expire_on_commit to ensure there is no old objects session.

davizucon: Another sugestion is to check if session avaliable, is yes reuse , if not create new one

gunthercox: Thank you, I've made a few modifications based on your feedback.

PR #748, opened 2017-05-18
Convert to Python3

ralic:
using : find . -name '*.py' | xargs 2to3 -w


gunthercox:
Hi @ralic thank you for this pull request. It look like the tool you used made a number of changes that actually break ChatterBot in both Python 2 and 3.

Did you encounter any errors in Python 3 that motivated these changes?

ChatterBot designed to be compatible with both Python 2 and 3 and it has continuous integration that helps ensure this https://travis-ci.org/gunthercox/ChatterBot (however there is always a chance that something was missed in due to an oversight in the test cases)

PR #776, opened 2017-06-07
fixing training generates response in reversed order

andpark:
fixing training puts previous statement as a response of current statement.
next statement should be the response of the previous one.

### DIFF
--- a/chatterbot/trainers.py
+++ b/chatterbot/trainers.py
@@ -78,7 +78,7 @@ class ListTrainer(Trainer):
         """
         statement_history = []
 
-        for text in conversation:
+        for text in reversed(conversation):
             statement = self.get_or_create(text)
 
             if statement_history:


chatterbot/trainers.py
```diff
@@ -78,7 +78,7 @@ def train(self, conversation):
         """
         statement_history = []
 
-        for text in conversation:
+        for text in reversed(conversation):
```
vkosuri: @andpark Thanks for contributing, The train reverse will not work for all. Instead you could make hack like this, If command line argument has train method you could reverse.

``` Python
if 'train' in sys.argv:
    conversation = reversed(conversation)
```

however this is not a good method, instead you could try a better solution for example detect Django command type 

for example
``` Python
if django.command.type == "train"
    "do your operations"
```

gunthercox:
@andpark Thank you for creating this pull request. It appears that the root of this issue was in ChatterBot's storage adapter which is used by the chat bot training module. I've opened a pull request to correct the storage adapter logic.

PR #787, opened 2017-06-10
Create SQL tables if they do not already exist

### LINKED ISSUE #741: Create SQL tables automatically if none exist
Currently it is required for `create=True` to be set when using to `SQLAlchemyDatabaseAdapter` for the first time so that it creates the tables in the database. This should only need to happen once.

- A check should be added so that when the storage adapter is initialized it checks to see if the tables exist.
- If the tables do not exist, then they should be created.
- The `create=True` parameter should be removed.

gunthercox:
Closes #741

### DIFF
--- a/chatterbot/storage/sql_storage.py
+++ b/chatterbot/storage/sql_storage.py
@@ -100,9 +100,7 @@ class SQLStorageAdapter(StorageAdapter):
             "read_only", False
         )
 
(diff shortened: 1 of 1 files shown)


vkosuri:
The build failed due to notwork issue, i have restarted build, i think it will pass

PR #809, opened 2017-06-30
Clean up training code

gunthercox:
This makes the following changes to the code in the training module:

- Variables have been renamed to be more accurate and descriptive.
- The ChatterBotCorpus trainer is no longer dependent on the ListTrainer.
- The entire statement history for a conversation is no longer stored (only the last statement is ever needed).

### DIFF
--- a/chatterbot/trainers.py
+++ b/chatterbot/trainers.py
@@ -76,17 +76,17 @@ class ListTrainer(Trainer):
         Train the chat bot based on the provided list of
         statements that represents a single conversation.
         """
-        statement_history = []
+        previous_statement_text = None
 
         for text in conversation:
             statement = self.get_or_create(text)
 
-            if statement_history:
+            if previous_statement_text:
                 statement.add_response(
-                    Response(statement_history[-1].text)
+                    Response(previous_statement_text)
                 )
 
-            statement_history.append(statement)
+            previous_statement_text = statement.text
             self.storage.update(statement)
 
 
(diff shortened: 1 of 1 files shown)


chatterbot/trainers.py
```diff
@@ -76,17 +76,17 @@ def train(self, conversation):
...
 
-            if statement_history:
+            if previous_statement_text:
                 statement.add_response(
-                    Response(statement_history[-1].text)
+                    Response(previous_statement_text)
                 )
 
-            statement_history.append(statement)
+            previous_statement_text.append(statement.text)
```
gunthercox: Should have been a equal sign :1st_place_medal:

chatterbot/trainers.py
```diff
@@ -102,20 +102,32 @@ def __init__(self, storage, **kwargs):
...
+                corpus_paths = corpus_paths[0]
 
         # Train the chat bot with each statement and response pair
-        for corpus in corpora:
-            corpus_data = self.corpus.load_corpus(corpus)
-            for data in corpus_data:
-                for pair in data:
-                    trainer.train(pair)
+        for corpus_path in corpus_paths:
+
```
vkosuri: @gunthercox Master, How about list comprehension, iterating each list will take long time if we train with huge data

gunthercox: List comprehension is always an option to improve performance. I do have concerns about readability though as the format of list comprehension is often difficult to understand at a quick glance.

Issue #815, opened 2017-07-02
Bottom to top approach for list comprehension


vkosuri:
With Referencing pull request comment https://github.com/gunthercox/ChatterBot/pull/809#discussion_r125018676

Yes, I agree with @gunthercox comment, readability :eyes: is very important it will gives us many advantages for example

- Catching issues will be very easier and new develop/integration will become very easier.
- Etc.

Yes sometime it is necessary to pick response also, With some hacks the below formats are looking good to me, the method i named it here :arrow_heading_up: **Bottom to top approach for list comprehension**, will provide more readability :eyes:

I did some exercise on some examples, please provide your comments/suggestion on this.

```Python
# Find the closest matching known statement
for statement in statement_list:
    confidence = self.compare_statements(input_statement, statement)

    if confidence > closest_match.confidence:
        statement.confidence = confidence
        closest_match = statement
```

vs

```Python
# The bottom to top approach for list comprehension, gives more readability
closest_match_and_confidence = [
    [
        statement,
        self.compare_statements(input_statement, statement)
    ] for statement in statement_list
]
```

Some more examples
```Python
for statement in statement_list:
    for response in statement.in_response_to:
        responses.add(response.text)
```
vs
```Python
# The bottom down approach for list comprehension, gives more readability
[
    responses.add(response.text)
    for statement in statement_list
    for response in statement.in_response_to
]
````

If it is looks good I'll try to make a PR which discussed here https://github.com/gunthercox/ChatterBot/pull/738

gunthercox:
@vkosuri I agree that this formatting does improve the readability of the list comprehension. Feel free to create a pull request. :+1:

PR #821, opened 2017-07-03
Remove statment_hsitory from UbuntuCorpus

vkosuri:
This pull request is intent to improve training process for larger ``datasets``. 

1. Removed ``statment_history`` in ``UbuntuCorpusTrainer``
2. Ignore ``data`` Ubuntu corpus files dir
3. Import order

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,5 @@ examples/ubuntu_dialogs*
 .out
 venv*
 
-# inore Ubuntu corpus files
+# ignore Ubuntu corpus files
 data
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ examples/ubuntu_dialogs*
 .env
 .out
 venv*
+
+# inore Ubuntu corpus files
+data
--- a/chatterbot/trainers.py
+++ b/chatterbot/trainers.py
@@ -49,7 +49,6 @@ class Trainer(object):
 
     def _generate_export_data(self):
         result = []
-
         for statement in self.storage.filter():
             for response in statement.in_response_to:
                 result.append([response.text, statement.text])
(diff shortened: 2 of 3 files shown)


chatterbot/trainers.py
```diff
@@ -78,17 +77,25 @@ def train(self, conversation):
...
+            for text in conversation:
+                statement = self.get_or_create(text)
+
+                if previous_statement_text:
+                    statement.add_response(
+                        Response(previous_statement_text)
+                    )
+
+                previous_statement_text = statement.text
+                yield statement
```
gunthercox: Was there a performance increase after this change?
These changes are quite different from what was discussed in #815.
Generator functions are a bit different from list comprehension as they produce a "lazy list" that can be called later. I'm not sure if this actually changes anything since we are calling the generator right away in this function.

vkosuri: > Was there a performance increase after this change?

It seems to me both are same.

> These changes are quite different from what was discussed in #815.

I thought i will combine both pull requests into one, now i will remove yield statements, I will push some more commit what we have disucssed.

> Generator functions are a bit different from list comprehension as they produce a "lazy list" 

Any more information about ``lazy lists``

vkosuri:
@gunthercox Master, This PR intended to remove only statments_hisotry form UbuntuCorpus, I will submit another PR about list comprehension

Any comments/suggestions?

chatterbot/trainers.py
gunthercox: > Any more information about lazy lists

This response on Stack Overflow seems to sum up generators fairly well https://stackoverflow.com/questions/1756096/understanding-generators-in-python

vkosuri:
@gunthercox are we going remove UbuntuCorpusTrainer in near feature? If it is true, can I close this?

chatterbot/trainers.py
```diff
@@ -136,9 +135,6 @@ class TwitterTrainer(Trainer):
     """
     Allows the chat bot to be trained using data
     gathered from Twitter.
-
-    :param random_seed_word: The seed word to be used to get random tweets from the Twitter API.
-                             This parameter is optional. By default it is the word 'random'.
```
gunthercox: Is there any reason that this was removed?

chatterbot/trainers.py
```diff
@@ -158,7 +154,6 @@ def __init__(self, storage, **kwargs):
     def random_word(self, base_word):
         """
         Generate a random word using the Twitter API.
-
```
gunthercox: This blank line was intended.

.gitignore
```diff
@@ -26,3 +26,6 @@ examples/ubuntu_dialogs*
 .env
 .out
 venv*
+
+# inore Ubuntu corpus files
```
gunthercox: Spelling error. `inore` should be `ignore`.

chatterbot/trainers.py
vkosuri: I'll updae

.gitignore
vkosuri: I'll update

chatterbot/trainers.py
vkosuri: I though :param: docstring is used only for procedures, if it is required I'll update it also

Issue #811, opened 2017-06-30
Session creation is duplicated


bilalbayasut:
Hi friends, does anyone experienced this problem :

I tried to make the chatbot to be used by multiple user 

``` Python
class SystemSession(object):
    def __init__(self, **kwargs):
        self.agent_id = kwargs.get('agent_id')
        self.session_id = kwargs.get('session_id')
        self.default_session_id = kwargs.get('default_session_id')

    def process(self):
        existing_session = Database.find_one('system_session', {"agent_id": self.agent_id})
        if not existing_session:
            session_id = self.create()
        else:
            session_id = existing_session['session_id']
        return session_id

    def get_all(self):
        pass

    def create(self):
        default_session = self.default_session_id
        data = {
            "agent_id": self.agent_id,
            "session_id": default_session,
            "created_at": datetime.now(),
            "updated_at": datetime.now()
        }
        Database.insert(collection='system_session', data=data)
        return default_session
```
By using code above, when I do json post and retrieve the session_id, the session_id is duplicated, means that the code generates the same session_id instead of unique one.. any ideas ?

FYI for this code I use the self.bot.default_session.id_string , as it generates new session. but I just realized when I post from the same PC by passing email and utterance , the bot code generates identical session_id string which makes it duplicated

gunthercox:
@bilalbayasut The default session id will be the same for each instance of a chat bot. You will want to generate a new id for each "user session" (however you are tracking that) that is created.

bilalbayasut:
I've implemented the new session from ConversationSessionManager , but I got this error any ideas ?

  File "/home/user/.local/lib/python3.6/site-packages/chatterbot/filters.py", line 31, in filter_selection
    if session.conversation.empty():
AttributeError: 'NoneType' object has no attribute 'conversation'

vkosuri:
@bilalbayasut Any further updates?

bilalbayasut:
Ok I have used the conversationManager, but the weird thing is when I use the generate_response() function and add seession_id within it, somehow the generate_base_query is failed.. somehow it looks like the session_id is None

  File "/home/bilal/.local/lib/python3.6/site-packages/chatterbot/storage/storage_adapter.py", line 41, in generate_base_query
    self.base_query = filter_instance.filter_selection(chatterbot, session_id)
  File "/home/bilal/.local/lib/python3.6/site-packages/chatterbot/filters.py", line 31, in filter_selection

gunthercox:
[...]

PR #828, opened 2017-07-07
Replace ChatterBot's "sessions" with "Conversations"

gunthercox:
This pull request is currently a work-in-progress.

For #811 to create consistent sessions that are tracked in the database instead of in memory.
Closes #513 by replacing it.


chatterbot/storage/django_storage.py
```diff
@@ -138,6 +138,21 @@ def remove(self, statement_text):
         responses.delete()
         statements.delete()
 
+    def get_latest_response(self, conversation_id):
+        """
+        Returns the latest response in a conversation if it exists.
+        Returns None if a matching conversation cannot be found.
+        """
+        from django.apps import apps
```
vkosuri: ``from django.apps import apps`` are many places, instead can we import one time on top of the page?

chatterbot/ext/django_chatterbot/migrations/0008_remove_phrase.py
```diff
@@ -0,0 +1,50 @@
...
+        ),
+        migrations.AlterField(
+            model_name='statement',
+            name='text',
+            field=models.CharField(max_length=255, unique=True),
+        ),
+        migrations.DeleteModel(
+            name='Phrase',
+        ),
+    ]
```
gunthercox: - [ ] Remember to squash and re-generate these migrations.

chatterbot/chatterbot.py
```diff
@@ -96,6 +96,8 @@ def get_response(self, input_item, session_id=None):
         :rtype: Statement
         """
         if not session_id:
```
vkosuri: Is this approach will produce more redability?

```Python
if session_id is not None and self.default_session_id is not None:
    self.default_session_id = self.stoage_create_conversation()
```

chatterbot/storage/django_storage.py
gunthercox: Importing third party packages at the top of a file sometimes causes unexpected errors depending on how users install ChatterBot. The best way (that I have found) to avoid this is to protect the imports by placing them inside of a method as this prevents them from being imported when the file is loaded. I'm actually not certain why some people encounter the import errors but it seems that it may be the result of less conventional (but still somewhat valid) installation methods such as installing directly through `setup.py` commands.

vkosuri:
💯

PR #837, opened 2017-07-11
Add warning about Ubuntu dialog corpus

gunthercox:
This adds a warning message in the documentation due to the performance issues when training with the Ubuntu dialog corpus.

CC @peterel


gunthercox:
@peterel Good catch. The change has been made.

PR #841, opened 2017-07-14
Reenable caching for MathematicalEvaluation

boidolr:
Whenever there was a cache hit the cache was subsequently reset.

### DIFF
--- a/chatterbot/logic/mathematical_evaluation.py
+++ b/chatterbot/logic/mathematical_evaluation.py
@@ -77,9 +77,7 @@ class MathematicalEvaluation(LogicAdapter):
 
         # Use the result cached by the process method if it exists
         if input_text in self.cache:
-            cached_result = self.cache[input_text]
-            self.cache = {}
-            return cached_result
+            return self.cache[input_text]
 
         # Getting the mathematical terms within the input statement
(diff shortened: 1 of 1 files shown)


chatterbot/logic/mathematical_evaluation.py
```diff
@@ -77,9 +77,7 @@ def process(self, statement):
 
         # Use the result cached by the process method if it exists
         if input_text in self.cache:
-            cached_result = self.cache[input_text]
-            self.cache = {}
-            return cached_result
+            return self.cache[input_text]
```
gunthercox: This change might cause a problem because it will keep caching statements forever until the program is restarted or the computer runs out of memory.

boidolr: Correct. Although the same is true of the previous version. It was consuming additional memory until some calculation was redone.
A different version might use `weakref` or `functoolslru_cache` (in python3) which both would not suffer this drawback.
Anyway, feel free to close the PR if you are not comfortable with the change.

gunthercox: Thank you. I'm going to close this off. I greatly appreciate the pull request though.

PR #848, opened 2017-07-16
Test if sudo=false reduces test time

gunthercox:
https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 
 python:


.travis.yml
```diff
@@ -1,3 +1,4 @@
+sudo: false
```
vkosuri: Still we are using libssl https://github.com/travis-ci/travis-ci/issues/3976 and mango are running through super user.

``` YAML
addons:
  apt:
    packages:
      - openssl

services:
  - mongodb
```

gunthercox: I looks like the containerized service that Travis uses for sudo:false builds doesn't require super privileges. I'm not certain of their implementation details but the tests are now passing.

gunthercox:
I'm going to close this off. Unfortunately it appears that this adds a significant amount of time to each build (~20 minutes).

PR #927, opened 2017-08-13
CONVERSATION_ID = bot.storage.create_conversation()

lesleslie:
so, default_conversation_id == None in chatterbot.py.

this causes a problem because when a query is executed on a sql database with a default_conversion_id of None it will produce an error:

DEFAULT_CONVERSATION_ID = bot.default_conversation_id
statement, response = bot.generate_response(input_statement, DEFAULT_CONVERSATION_ID)

will cause an error in line 261 of sql_storage.py:

conversation = session.query(Conversation).get(conversation_id)

because converation_id cannot be equal to None you will get the following:

TypeError: object of type 'NoneType' has no len()

What I propose is this:

CONVERSATION_ID = bot.storage.create_conversation()

this change makes more sense to me because each time you access the bot via terminal, or however you're doing it, it actually is it's own conversation and should be treated as such. let's say you talk to your bot in terminal mode and then somebody else then talks your bot in terminal mode - these are two separate conversations and should be treated as such.  i've tested this with sqlite and mongodb it seems to work just fine.

along this same line of thought then we'll change:

bot.storage.add_to_conversation(bot.default_session, statement, response)

to:

bot.storage.add_to_conversation(CONVERSATION_ID, statement, response)

i'm also adding:

bot.storage.add_to_conversation(CONVERSATION_ID, statement, response1)

to learning_new_response.py.  this way the new response will be saved to the conversation too.

-- les

### DIFF
--- a/chatterbot/chatterbot.py
+++ b/chatterbot/chatterbot.py
@@ -115,7 +115,7 @@ class ChatBot(object):
         previous_statement = self.storage.get_latest_response(conversation_id)
 
         if not self.read_only:
-            self.learn_response(statement, previous_statement)
+            self.learn_response(response, previous_statement)
             self.storage.add_to_converation(conversation_id, statement, response)
 
         # Process the response output with the output adapter
--- a/chatterbot/chatterbot.py
+++ b/chatterbot/chatterbot.py
@@ -116,7 +116,7 @@ class ChatBot(object):
 
         if not self.read_only:
             self.learn_response(statement, previous_statement)
-            self.storage.add_to_converation(conversation_id, statement, response)
+            self.storage.add_to_conversation(conversation_id, statement, response)
 
         # Process the response output with the output adapter
         return self.output.process_response(response, conversation_id)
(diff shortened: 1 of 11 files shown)


radiofrequency:
File "learning_feedback_example.py", line 26, in <module>
    CONVERSATION_ID = bot.storage.create_conversation()
AttributeError: 'SQLStorageAdapter' object has no attribute 'create_conversation'

gunthercox:
@radiofrequency I believe that error message was the result of a spelling issue in a method name. It has since been corrected.

yollychen:
i used the commend to installed chatterbot
`pip install chatterbot`
but it still have this error
AttributeError: 'SQLStorageAdapter' object has no attribute 'create_conversation'
and i found if i used commend to install , i got different code on git

vkosuri:
@yollychen I think these changes in master not yet pushed into pip, try to use GitHub master version to resolve this issue.

yollychen:
@vkosuri cause i use virtual environment, so i only can use pip to install. When will it pushed into pip?

PR #941, opened 2017-08-16
Extra data is a field in Django that will disappear sooner or later, …

Issen007:
…so this field is not required anymore when creating a new Statement/responds manually in Django Admin.

### DIFF
--- a/chatterbot/ext/django_chatterbot/models.py
+++ b/chatterbot/ext/django_chatterbot/models.py
@@ -16,7 +16,10 @@ class AbstractBaseStatement(models.Model):
         max_length=255
     )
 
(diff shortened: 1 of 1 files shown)


Issen007:
@gunthercox

gunthercox:
Does this fix anything or is this just an ascetic change?

Issen007:
@gunthercox this is only a smaller change, because in existing version if I try to create a Statement or Responds manually today, then is the `Extra Data` field a required field that you must fill out.

With this smaller change I have change that column to no mandatory field so you can still create a statement or responds manually without to fill out the `Extra data` field.

Issue #925, opened 2017-08-12
Allow statements to be filtered by their tags


gunthercox:
_This issue is blocked by #924 and can't be worked on until those changes are made._

***

Changes need to be made to ChatterBot's storage adapters so that statements can be filtered by their `tags` attribute.

This is an important change that will allow more efficient response generation. After this change is made it will be possible to add a processing step where new input statements are classified based on the labels of categories of the dialog that was used to train the chat bot (adding this additional processing step will have a separate issue and pull request created for it).

vkosuri:
@gunthercox Master, If you didn't started working on this, I am happy to work on this feature, Is there any ETA to do this?

vkosuri:
I am planning to write a utility method to find categories something like below

Question: If same/similar statement present in multiple categories are we going considering those?

``` Python
def find_categories(corpus_paths, input_statement):
    data_file_paths = self.list_corpus_files(corpus_paths)
    categories = [list]
    for file_path in data_file_paths:
        data = read_corpus(file_name):
        for k, v in data.items():
            if input_statement in v:
                categories.append(k)
                
    return categories
```

gunthercox:
@vkosuri Thank you, that would be greatly appreciated.

Yes, multiple categories for a statement should be supported.

Also, if you check the `corpora` variable on [#L119](https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/trainers.py#L119) of trainers.py you should be able to get the categories from it. They are being added in the [`load_corpus`](https://github.com/gunthercox/chatterbot-corpus/blob/master/chatterbot_corpus/corpus.py#L77) function of `chatterbot-corpus`.

```
corpora = self.corpus.load_corpus(corpus_path)
for category in corpora.categories:
    # ...
```

Thank you again, let me know if anything comes up that I can assist with. 👍

vkosuri:
Thanks Master, Another question

1. is ``None`` is the default category?
2. If user didn't specified any category about do we need search whole corpora? If not what's your thoughts?

gunthercox:
> 1. is None is the default category?

Since a statement can have multiple categories, the default should be an empty list.

> 2. If user didn't specified any category, do we need search whole corpora? If not what's your thoughts?

Yes, but that is going to be a different pull request. #925 describes this a bit more.

Issen007:
@gunthercox @vkosuri in what area should we start using tags?
I'm playing around with this at the moment and could I add tags example in to the Corpus training data files?

[...]

PR #945, opened 2017-08-18
Add inverted index for mongodb

mymusise:
Signed-off-by: corcassia <mymusise1@gmail.com>
Hi, I add the Inverted Index for mongodb. What new:
- Creating the Inverted Index when training dataset
- It will just compare the similar statement when getting a response (Only `BestMatch`)

Improving performance：

| Version | Response Time(s) |
| ---------- | ------------------------ |
| Before | 3.1 |
| After | 0.1 |

It test with 28000 rows dataset on Ubuntu 14.04 + Core i5 4670 + 12G Menery.

Hope it helps.


vkosuri:
It seems to me the build failed due to network issue, i have restarted I have restarted the job, hope it will pass

corcassia:
@gunthercox  Hi, it's strange that the CI failed with `zipfile.BadZipfile: File is not a zip file` in `#2677.1`.
But I run `tox` pass all test in my python2.7 environment on Ubuntu 14.04.
Have you meet it before?

vkosuri:
Some useful reference about invert index https://stackoverflow.com/a/8360932/358458

vkosuri:
@corcassia Thank you for you PR, could you please clarify the different between [mongodb text indexing](https://docs.mongodb.com/v3.2/core/index-text/) and this PR invert indexing

corcassia:
@vkosuri  Ok, my pleasure.

Before this PR, I found it will compare all samples when calling `ChatBot.get_response`. So, if we have trained a large dataset, it will spend many time to response.
`mongodb text indexing` is a `forward index`, in this change, I hope it just select the similar sample from database by `invert indexing`, rather than select all.

For example:

vkosuri:
@corcassia Thank you, I am not sure with performance. 

To improve @gunthercox came with an idea **tag filtering** to speed up retrieval process https://github.com/gunthercox/ChatterBot/issues/925.

corcassia:
@vkosuri  Yes, I think **tag filtering** is a good idea. 
But, to get the tag of input statement, will it iterate all sample or pass a **classify model**?

gunthercox:
@corcassia Thank you, I'm really impressed with the performance improvements after this change. I'm going to pull these changes down locally and test out a few things. I might have further questions.

gunthercox:
@corcassia Apologies for my delay, I will be reviewing this pull request as soon as possible.

chatterbot/storage/mongodb.py
```diff
@@ -331,27 +381,43 @@ def remove(self, statement_text):
...
+            _statement_query = {
+                'text': {
+                    '$in': responses
+                }
+            }
+        else:
+            # Just filter the statement in need.
+            statement_ids = []
+            tokens = self.statement_segmentation(statement.text)
+            word_dicts = self.inverted.find({'word': {'$in': tokens}})
```
gunthercox: I might be misunderstanding this, so feel free to correct me if I'm wrong. Given this change, wont only statements in which the words of `statement` exist be returned?

corcassia: Yes, you're right. So this change will lead to no response if there's no statement exist.

corcassia:
@gunthercox It's all right, and I found some problem with this change in these days.

corcassia:
@vkosuri Hey, I was wrong. `MongoDB text indexing` is one of the practices of `invert indexing`.  After a few test, I found `MongoDB text indexing` behave better. 
I remove my `invert indexing` and change the `_statement_query`.

zxsimple:
@mymusise seems the change hasn't checked in to the master? can I apply the path master brach for resolving the performance issue?

mymusise:
@zxsimple Yeah, seems the master branch still not create a full-text index when setting up the bot [(chatterbot/storage/mongodb.py)](https://github.com/mymusise/ChatterBot/blob/master/chatterbot/storage/mongodb.py#L115), but you can create it manually and with a little change [like this](https://github.com/mymusise/ChatterBot/blob/master/chatterbot/storage/mongodb.py#L385).

If changing source code is a boresome work, there's another way to resolving the performance.

I think the performance with highly improve with decreasing the [`statement_list`](https://github.com/mymusise/ChatterBot/blob/master/chatterbot/logic/best_match.py#L35), cause [the comparison](https://github.com/mymusise/ChatterBot/blob/master/chatterbot/logic/best_match.py#L36) will cost a lot of CPU time.I found it may help by removing some meaningless works such as ('the', 'a', 'an', 'with', 'to') from input statement and corpus.

## Solution:
```python
import re

source_corpus = [
    "Hello",
    "Hi there!",
    "How are you doing?",
    "I'm doing great.",
    "That is good to hear",
    "Thank you.",
    "You're welcome."
]
input_statements = [
    "Such a good day!",
    "Here we are."
]

def words_filter(statements):
    # a words filter demo
    bad_words = ['the', 'a', 'an', 'is', 'are']
    rule = re.compile(" | ".join(bad_words))
    return [rule.sub(' ', statement) for statement in statements]

courpus = words_filter(source_corpus)
input_statements = words_filter(input_statements)

from chatterbot.trainers import ListTrainer
from chatterbot import ChatBot

chatbot = ChatBot("Ron Obvious")

chatbot.set_trainer(ListTrainer)
chatbot.train(courpus)


for statement in input_statements:
    response = chatbot.get_response(statement)
    print(response)
```

zxsimple:
@mymusise ， thanks for your reply, I tried and found the bottleneck related to the number of question dataset. I have 3K+ questions, each of the comparison takes 1ms, so each response to the question takes around 3 seconds. The complexity of comparison is O(n), I'll try KNN for similarity comparison and see. 

`        for statement in statement_list:
            confidence = self.compare_statements(input_statement, statement)
            if confidence > closest_match.confidence:
                statement.confidence = confidence
                closest_match = statement`

PR #968, opened 2017-09-01
Fix typo

chellem:


gunthercox:
Hi @chellem thank you for the pull request. The title of this pull request is a bit misleading as this appears to add new documentation, not fix a typo.

The documentation being added appears to address the json storage adapter which is deprecated and is going to be removed from ChatterBot soon. I'm going to decline these changes because the json storage adapter was intentionally undocumented to encourage developers not to use it any more.

PR #988, opened 2017-09-20
Getting recent response from the database.

kishore-devaraj:
Hi,

I had a use case where i need to get the latest response of the same message. For instance, for the text 'hi' there are two responses 'hello' and 'hey' in the database. What i want is, I need to get the response which has been included recently (say 'hey' in this example). 

I checked out the response_selection  method and it seems to have get_first_response and other kind of methods but this one is not included. So i created a new method called get_recent_response in chatterbot.response_selection.py file and it seems to fulfill my use case. 

I thought may be we can add this to our chatterbot module. So this is the reason why i created this pull request.  Please checkout the code and if any queries feel free to ask. 

Thank you.

### DIFF
--- a/chatterbot/response_selection.py
+++ b/chatterbot/response_selection.py
@@ -51,6 +51,23 @@ def get_first_response(input_statement, response_list):
     ))
     return response_list[0]
 
+def get_recent_response(input_statement,response_list):
+    """
+    :param input_statement: A statement, that closely matches an input to the chat bot.
+    :type input_statement: Statement
+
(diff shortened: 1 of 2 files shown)


gunthercox:
Hi @kishore-devaraj, thank you for your interest in contributing to this project. This pull request looks great, your code is very clean, you've provided a use case, and unit tests.
Unfortunately I'm hesitant to merge this pull request in because the function you've added isn't used by any other parts of ChatterBot. In the interest of keeping this package lean and maintainable, I'm going to decline this pull request.

PR #1021, opened 2017-10-11
Add tagging when training with SQL storage

### LINKED ISSUE #924: Modify ChatterBot Corpus trainer to save corpus categories as tags for statements
This is a small change that needs to be made before future changes that will help improve the time and processing efficiency when ChatterBot generates a response to an input statement.

Changes have already been made to add categories to the [ChatterBot Corpus](https://github.com/gunthercox/chatterbot-corpus/) so that the data looks like this:

```yaml
categories:
- conversations
conversations:
- - Good morning, how are you?
  - I am doing well, how about you?
  - I'm also good.
  - That's good to hear.
  - Yes it is.
- - Hello
  - Hi
  - How are you doing?
  - I am doing well.
```

The `categories` attribute allows a list of descriptive labels to be added to the collection of dialog. Now, changes need to be made so that the `ChatterBotCorpusTrainer` in [trainers.py](https://github.com/gunthercox/ChatterBot/blob/master/chatterbot/trainers.py) adds the values from the `categories` list in the corpus data to the `tags` fields that statements have.

***

In the future (a different issue and pull request will be created to describe and make this change), the categorization of statements will make it possible for the chat bot to predict a smaller set of data that a possible response should be likely to exist it.

gunthercox:
Closes #924

- [x] Tests

### DIFF
--- a/chatterbot/conversation/statement.py
+++ b/chatterbot/conversation/statement.py
@@ -2,7 +2,27 @@
 from .response import Response
 
 
-class Statement(object):
+class StatementMixin(object):
+    """
+    This class has shared methods used to
+    normalize different statement models.
+    """
+
+    def get_tags(self):
+        """
+        Return the list of tags for this statement.
+        """
+        return self.tags
+
+    def add_tags(self, tags):
+        """
+        Add a list of strings to the statement as tags.
+        """
+        for tag in tags:
+            self.tags.append(tag)
+
+
+class Statement(StatementMixin):
     """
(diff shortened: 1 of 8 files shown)


kyle-sorensen:
This is the pull request to add tags to statements, which will greatly increase efficiency on trained models, right?

gunthercox:
@kyle-sorensen It will make it possible to do that. Other pull requests will need to be created after this one that apply filtering by tags database queries that retrieve possible response candidates.

PR #1043, opened 2017-10-28
Support for Django 2.0

rajasimon:
From Django 2.0 Release [note](https://docs.djangoproject.com/en/2.0/releases/2.0/#features-removed-in-2-0)...

> The on_delete argument for ForeignKey and OneToOneField is now required in models and migrations. Consider squashing migrations so that you have less of them to update.

So on_delete must needed in ForeignKey so update to Django 2.0 won't break the development.


gunthercox:
@rajasimon Thank you for the pull request. I'm going to check the following two things and if everything checks out, I'll pull this in.

- [ ] Check if database migrations need to be created
- [ ] Double check that these changes are still compatible with Django < 2.0

PR #1057, opened 2017-11-06
Add twitter_lang param to TwitterTrainner

tentamen:
I think it's nice to have a param to specific tweets language when gathering data from Twitter.


gunthercox:
@tentamen Thank you for your contribution. I agree that this parameter is a useful addition.

tentamen:
Glad I could help

PR #1121, opened 2017-12-16
Time logic adapter improvements

### LINKED ISSUE #994: Why using NLTK Naive Bayes Classifier used in Time logic adapter?
[NLTK Naive Bayes Classifier](http://www.nltk.org/_modules/nltk/classify/naivebayes.html) only return 1 or 0 as confidence and it always return 1 wrongly when user input several characters(e.g abc), 

https://github.com/gunthercox/ChatterBot/blob/a4a7b0287c3516975770d6f9ab1e5b8fa5572e1c/chatterbot/logic/time_adapter.py#L61

why don't you use compare_statements in best match adapter which return a proper confidence and more standardized. 

https://github.com/gunthercox/ChatterBot/blob/351971a21e603642f5797c33e924c4195f96682b/chatterbot/logic/best_match.py#L36

gunthercox:
This adds the ability to pass in custom positive and negative
training data for the NaiveBayesClassifier.

A few new positive and negative data elements have been added.

A first word feature has been added to help identify questions.

Closes #994

### DIFF
--- a/chatterbot/logic/time_adapter.py
+++ b/chatterbot/logic/time_adapter.py
@@ -6,33 +6,47 @@ from .logic_adapter import LogicAdapter
 class TimeLogicAdapter(LogicAdapter):
     """
     The TimeLogicAdapter returns the current time.
+
+    :kwargs:
+        * *positive* (``list``) --
+          The time-related questions used to identify time questions.
+          Defaults to a list of English sentences.
+        * *negative* (``list``) --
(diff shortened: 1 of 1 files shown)


nggary:
I am afraid that the NaiveBayesClassifier fail to classify the input string in TimeLogicAdapter, I have tried to input "Who are you" & "How are you" some greeting sentence, the confidence always return 1, not to mention if I implemented multiple adpater.

PR #1141, opened 2018-01-06
Added persona object.

madhupjain:


gunthercox:
Hi @madhupjain, thank you for the pull request. I'm not sure if there is any benefit to merging in this change though.

Issue #939, opened 2017-08-16
Removing support for redundant database connection parameters


gunthercox:
Right now, the MongoDB storage adapter and the SQL storage adapter each support both of the following parameters.

- `database`: A simple string such as "chat-bot-database" that then gets turned into a database connection string by the storage adapter.
- `database_uri` A database connection string.

I think it might simplify things if one of these parameters was removed. Considering they both almost do the same thing I don't think there would be too much of an issue if `database` was removed completely as a possible parameter for these two adapters.

***

> "Explicit is better than implicit."
> ~ [The Zen of Python](https://www.python.org/dev/peps/pep-0020/)

dapp1990:
I'm working on this issue. Can I take it?

gunthercox:
Yes, please do. Thank you :smile:

PR #1161, opened 2018-01-18
Removing support for redundant database connection parameters (#939)

dapp1990:
Now database parameter is not longer supported by MongoDB and SQL adapters.

In the case of MongoDB, database_name was kept as local variable to have a clean reference for datase other than the pure database_uri (to avoid non-alphanumeric characters in file name).

### DIFF
--- a/chatterbot/storage/mongodb.py
+++ b/chatterbot/storage/mongodb.py
@@ -1,5 +1,4 @@
 from chatterbot.storage import StorageAdapter
-import re
 
 
 class Query(object):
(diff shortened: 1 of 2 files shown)


dapp1990:
I've tried to fix the issue on travis but I'm not sure what could be the problem. Do you have any idea about this issue?

´ ERROR: InvocationError: '/home/travis/build/gunthercox/ChatterBot/.tox/test/bin/nosetests --with-coverage --cover-package=chatterbot' ´

mikekeda:
Check the tests, some of them are failed.
Try to run tests locally.

PR #1192, opened 2018-01-28
Fix storage methods for getting the latest response

gunthercox:
- [ ] Remove all `@expectedFailure` decorators and have tests pass.

PR #1199, opened 2018-02-03
fix mongo storage get_latest_response fun for the first result

yangl:


yangl:
https://github.com/gunthercox/ChatterBot/pull/1192

PR #1207, opened 2018-02-13
Fix "Statement matching query does not exist" error

mayankkushal:
When using a logic adapter, if an unknown statement was given in Django project,  `chatterbot.ext.django_chatterbot.models.DoesNotExist: Statement matching query does not exist.` error used to pop up.  Happens in `chatterbot.logic.LowConfidenceAdapter` for default response and the other logic adapters as well.

### DIFF
--- a/chatterbot/storage/django_storage.py
+++ b/chatterbot/storage/django_storage.py
@@ -181,8 +181,8 @@ class DjangoStorageAdapter(StorageAdapter):
         Statement = self.get_model('statement')
         Response = self.get_model('response')
 
-        first_statement = Statement.objects.get(text=statement.text)
-        first_response = Statement.objects.get(text=response.text)
+        first_statement, created = Statement.objects.get_or_create(text=statement.text)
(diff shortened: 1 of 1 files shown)


chatterbot/storage/django_storage.py
```diff
@@ -181,8 +181,8 @@ def add_to_conversation(self, conversation_id, statement, response):
         Statement = self.get_model('statement')
         Response = self.get_model('response')
 
-        first_statement = Statement.objects.get(text=statement.text)
-        first_response = Statement.objects.get(text=response.text)
+        first_statement, created = Statement.objects.get_or_create(text=statement.text)
```
vkosuri: But the changes are looks good, but why you are storing into another variables? Could please elaborate?

mayankkushal: Do you mean about `created` ? Its because `get_or_create` returns a tuple of (object, created). No apparent use of `created` but still need keep it there to separate out the variables.

PR #1289, opened 2018-04-29
Datetime parsing does recognize am or pm

vkosuri:
@coolrb The script assumes 24hr format by default, to work for you scenario you have to write like this

Scenario 1

```Python
from chatterbot import parsing
text= u'Your dental appointment with Dr P. Delvour is scheduled for October 29, 4:00pm. ABC Dentist, 555-555-555.'
dates_ch = parsing.datetime_parsing(text)
print(dates_ch)
```
Update regular expression
```
re_time = r'(?P<hour>\d{1,2})(\:(?P<minute>\d{1,2})\s?(?P<convention>am|pm))'
```

Scenario 2:
```Python
from chatterbot import parsing
text= u'Your dental appointment with Dr P. Delvour is scheduled for October 29, 16:00. ABC Dentist, 555-555-555.'
dates_ch = parsing.datetime_parsing(text)
print(dates_ch)
```
Scenario 3:
Skip minutes, the regex will work.

```Python
from chatterbot import parsing
text= u'Your dental appointment with Dr P. Delvour is scheduled for October 29, 4 pm. ABC Dentist, 555-555-555.'
dates_ch = parsing.datetime_parsing(text)
print(dates_ch)
```
Closes: https://github.com/gunthercox/ChatterBot/issues/1277

PR #1291, opened 2018-05-01
The parser just recognizes case sensitive strings

manoelrui:
@gunthercox I tested the parser with some test cases to get **datetime** objects and I realised that upper letters are crashing the parser when it try to compare strings. 

In the first scenario below, as the parser uses the function **date_from_relative_week_year**, the function argument **time** assigned with string "Next" is not equals to string "next" and the function will return a "NoneType".

_Scenario 1:_
```
from chatterbot import parsing
text = "Next Week"
parser = parsing.datetime_parsing(text)
print(str(parser))

```
In the second scenario, the parser uses the function **convert_time_to_hour_minute**. As similar before, the function will return a "NoneType" because the comparison between the **convention** argument (storing the string "PM") and string "pm" is false. 

_Scenario 2:_
```
from chatterbot import parsing
text = "2PM"
parser = parsing.datetime_parsing(text)
print(str(parser))
```

### DIFF
--- a/chatterbot/parsing.py
+++ b/chatterbot/parsing.py
@@ -206,8 +206,8 @@ regex = [
         ),
         lambda m, base_date: date_from_relative_week_year(
             base_date,
-            m.group('time'),
-            m.group('dmy'),
+            m.group('time').lower(),
+            m.group('dmy').lower(),
             m.group('number')
         ) + timedelta(**convert_time_to_hour_minute(
             m.group('hour'),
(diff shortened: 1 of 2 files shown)


gunthercox:
@manoelrui I think you've correctly identified this bug. Would you be able to add some test cases, similar to what @vkosuri had in his [similar pull request](https://github.com/gunthercox/ChatterBot/pull/1289/files)?

gunthercox:
@manoelrui The test cases look great, awesome work!

gunthercox:
I've merged the commits 64762ba6fd58ed2057e7c620851c3bd93185f65a dbcda9a3b710d777d683492e035ab30ef5bcce39 into master by cherry-picking them. The changes will be available in the next ChatterBot release. @manoelrui Thank you again.

Issue #1294, opened 2018-05-03
Reverse answers with Django again?


vpena:
Hello!

I've been trying to use Chatterbot with Django and I'm getting reversed answers, which does not happens when I use Chatterbot with a python script.

To reproduce:
1) Clone Chatterbot from github
2) Go to examples, django_app and pip install requirements.txt with python 3.5.2 (getting v0.8.5 of Chatterbot)
3) Change to this configuration:
``` Python
CHATTERBOT = {
    'name': 'Django ChatterBot Example',
    'trainer': 'chatterbot.trainers.ChatterBotCorpusTrainer',
    'training_data': [
        'chatterbot.corpus.english.gossip'
    ],
    'django_app_name': 'django_chatterbot'
}
```
Try this conversation:
Me: "did tell gossips to anybody"
Answer: "Jimmy said he give order to ai's like I and they do as he say." **(This is ok)**
Me: "tell me what you know about gossips"
Answer: "John said he follow a saying to get a friend I must be a friend." **(This is reversed answer)**

If I try the same without django:
``` Python
from chatterbot import ChatBot

chatbot = ChatBot(
    'Charlie',
    trainer='chatterbot.trainers.ChatterBotCorpusTrainer',
    input_adapter="chatterbot.input.TerminalAdapter",
    output_adapter="chatterbot.output.TerminalAdapter",
    database="../test_corpustrainer.db",
)

chatbot.train(
    'chatterbot.corpus.english.gossip'
)
while True:
    try:
        chatbot_input = chatbot.get_response(None)
    except (KeyboardInterrupt, EOFError, SystemExit):
        break
```
Me: "did tell gossips to anybody"
Answer: "Jimmy said he give order to ai's like I and they do as he say."
Me: "tell me what you know about gossips"
Answer: "Jordan said he notice that lots of other people are talking to me how many." **(correct order of answer according to the corpus)**

Am I doing something wrong? 

Cheers and thank you very much,
Vanessa

vkosuri:
@vpena Thanks, It seems to me you haven't created pull request, could you please do so?

niranjan8129:
how to execute the program  ... what is the main source file to pass the text

vkosuri:
@niranjan8129 You have install all the required packages into machine, ``python manage.py train`` and ``python manage.py runserver``

About Django the documentation found here https://chatterbot.readthedocs.io/en/stable/django/index.html

[...]

PR #1295, opened 2018-05-04
Attempt to fix https://github.com/gunthercox/ChatterBot/issues/1294

sfehlandt:
Trying to solve this use-case: https://github.com/gunthercox/ChatterBot/issues/1294

We do not know how to test extensively if other use cases are still working properly. We will use this modified version, and we will let you know if we find any issue


vkosuri:
@sfehlandt Thanks for contributing, could you try to add some tests to resolve this issue in future.

gunthercox:
@sfehlandt Thank you for looking into this issue. I apologize that I never got around to reviewing this or providing any suggestions. I'm going to close this pull request due to the number of changes that have altered the way the Django storage adapter works since this pull request was made.

PR #1309, opened 2018-05-18
Fixed misguided comment

nihal111:
Fixed erroneous comment in basic_example.py
Arises from change made in ae4ad35


gunthercox:
@nihal111 Thank you.

PR #1359, opened 2018-07-30
Added alternative installation method

cheezeburger:
python setup.py install


gunthercox:
Hi @cheezeburger, I really appreciate you taking the time to make this contribution. I apologize for not responding to this pull request sooner. Although I would love to document alternate installation methods for ChatterBot, doing so adds an additional degree of difficulty for me when it comes to helping debug issues that people report. (As strange as it sounds, there have been differences between installing with pip and setup.py). I'd prefer there to be one, recommended method of installation because of this.

Issue #1126, opened 2017-12-19
IndexError in get_latest_response method for Mongo DB storage adapter


kennedyshead:
``` Bash
Traceback (most recent call last):
  File "/Users/magnusknutas/git/automation/jarvis/bot/__init__.py", line 70, in process_audio
    self.give_response(value)
  File "/Users/magnusknutas/git/automation/jarvis/bot/__init__.py", line 92, in give_response
    resp = self.chatbot.get_response(data)
  File "/Users/magnusknutas/git/automation/venv_jarvis/lib/python3.6/site-packages/chatterbot/chatterbot.py", line 116, in get_response
    previous_statement = self.storage.get_latest_response(conversation_id)
  File "/Users/magnusknutas/git/automation/venv_jarvis/lib/python3.6/site-packages/chatterbot/storage/mongodb.py", line 297, in get_latest_response
    return self.mongo_to_object(statements[-2])
IndexError: list index out of range
```

gunthercox:
Hi @kennedyshead Can you provide any instructions to help us recreate this error so that we can help you?

kennedyshead:
I think it was on a fresh install with no corpus. I added a check so that the statements list contained the amount of elements needed and that worked.

Wybxc:
I met the same problem.
I am Chinese and I use my own [corpus](https://github.com/Wybxc/chinese-corpus). But when I call `get_response` the second time, I get an `IndexError`. 
```
Traceback (most recent call last):
  File "..\chat.py", line 11, in <module>
    print(r(i))
  File "..\chat.py", line 6, in r
    def r(s):return bot.get_response(s).text
  File "E:\Anaconda3\lib\site-packages\chatterbot\chatterbot.py", line 122, in get_response
    previous_statement = self.storage.get_latest_response(conversation_id)
  File "E:\Anaconda3\lib\site-packages\chatterbot\storage\mongodb.py", line 297, in get_latest_response
    return self.mongo_to_object(statements[-2])
IndexError: list index out of range
```
I do not use other corpus.
And by disabling MongoDB it works well, but in that case it is too slow.
The problem only occurs on my new computer, which is Windows 10 x64. It works well on Windows 7 x86.

gunthercox:
I'm going to close this ticket off for now, the version of the `get_latest_response` has been moved to the `ChatBot` class as a part of the changes made in #1391. This should no longer be an issue.

PR #1372, opened 2018-08-16
FIX: IndexError in mongodb.py

Wybxc:
fix the bug in #1126 .


gunthercox:
The `get_latest_response` has been moved to the `ChatBot` class as a part of the changes made in #1391. Additional tests have been added to cover the new design.

@Wybxc thank you for the pull request, although this didn't end up being a solution the work you put into it is greatly appreciated.

PR #1391, opened 2018-09-12
Remove Conversation and Response objects to simplify data representations

gunthercox:
This pull request makes a very large change to the way conversations are represented in ChatterBot. Previously, a conversation was its own object that responses could be linked to. Now, a conversation is a label on statement and a response is a text field on the statement. In addition, the text attribute of statements no longer has a unique constraint.
***
The `find` method has also been removed from storage adapters. It functioned with the constraint that a statement's `text` is always unique, which is no longer the case.


vkosuri:
+1

PR #1420, opened 2018-09-22
Remove set_trainer method and trainer attribute from the ChatBot class

gunthercox:
This makes the following changes:

* Removes the `set_trainer` method from the `ChatBot` class.
* Removes the `trainer` attribute from the `ChatBot` class.

* Updates the documentation to show to initialize an instance of a training class with a ChatBot instance.


vkosuri:
+1

PR #1439, opened 2018-10-03
<swaraj> was  here

swaraj961:
added some emojis to the previous code


gunthercox:
Hi @swaraj961, I appreciate the contribution but I think I'd prefer to keep the README headings plain for now.

PR #1444, opened 2018-10-07
Fixed running example

rajendraarora16:
I have seen several folks were struggling to run first example since I am fixing this example and fortunately would work fine.


gunthercox:
Hi @rajendraarora16, thank you for the contribution.

I can't accept these code changes because they won't run on the latest (alpha) version of ChatterBot. The `set_trainer` function was removed in a previous pull request.

PR #1445, opened 2018-10-07
Filter out 'bot:' statements from possible responses

gunthercox:
This pull request will make changes so that statements that are generated by the chat bot will be excluded from the set of possible responses that the bot can respond with. This essentially means that the bot will "doubt" it's own ability to respond accurately and instead trust the statements that come from people that it communicates with.


chatterbot/storage/mongodb.py
```diff
@@ -227,6 +228,9 @@ def get_response_statements(self):
         _response_query = {
             'in_response_to': {
                 '$ne': None
+            },
+            'persona': {
+                '$not': re.compile('^bot:*')
```
gunthercox: I'm quite impressed that it's possible to pass a Python regular expression to Mongo DB. I'd be interested to know how this works on the database side of things.

PR #1452, opened 2018-10-13
Use stemmed text to search for statement matches

gunthercox:
Using the stemmed statement text should provide more efficient search and text comparison functionality.

### DIFF
--- a/chatterbot/stemming.py
+++ b/chatterbot/stemming.py
@@ -1,5 +1,6 @@
 import string
 from nltk import pos_tag
+from nltk.data import load as load_data
 from nltk.corpus import wordnet, stopwords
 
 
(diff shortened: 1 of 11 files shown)


chatterbot/stemming.py
```diff
@@ -0,0 +1,47 @@
...
+
+        words = []
+
+        # Generate the stemmed text
+        for word in text.split(' '):
+
+            # Remove stopwords
+            if word not in self.stopwords:
+                # Chop out the middle of the word
+                start = self.get_chop_size(len(word))
```
gunthercox: Chopping might not be required. Removing punctuation and stopwords might be enough to provide a simpler search index.

Some tests with existing data might provide more relevant support for this hypothesis.

PR #1488, opened 2018-11-13
Switch to bigram strategy for text indexing

gunthercox:
This change will reduce the size of the result matching set by decreasing the number of possible matches based on the need for similar text to have at least two neighboring matches.

For #1452


vkosuri:
More info about https://en.m.wikipedia.org/wiki/Bigram

PR #1529, opened 2018-12-12
Basic example fix #1426

MartinDelille:
This is an attempt to fix the error in `examples/basic example.py`.

It now runs but the answer to the question *I would like to book a flight.* is strange: *Much the same as being a human, except that we lack all emotions, dreams, aspirations, creativity, ambition, and above all subjectivity.*

It looks like it has been ask *Hello, how are you today?*

PR #1540, opened 2018-12-27
Use a class-level instance for the sentence detector

gunthercox:


gunthercox:
@MartinDelille I was able to recreate the issue you described in #1529 related to `punkt` not being downloaded. I've made a change in this pull request which ensures that the `punkt` data is downloaded before the sentence detector is used.

PR #1543, opened 2018-12-31
Separate search process from logic adapters

gunthercox:
Once completed, this pull request will make changes to separate the search process (currently the `get` method in logic adapters) into it's own module. This change will allow more defined search behavior in the future, and also presents a better organization that would allow caching of search results to be used in the future to improve response times.


vkosuri:
:100:

PR #1590, opened 2019-01-28
updated the init and process signatures

ryancollingwood:
When using the example code was getting an error of too few arguments, updated the code based on what I saw in the built in logic adapters


gunthercox:
@ryancollingwood Thank you.

Issue #1203, opened 2018-02-08
Weird results in the chatterbot.chatterbot:Adding responses


jasonleehodges:
The logs are saying something like 'Adding "Hey" as a response to "Hey"' which for the first iteration here makes sense but subsequent iterations it keeps saying that it's adding the response to the phrase "Hey" even thought that was not the last response. 

For example, further down the chain you will see that the bot says "See you later" to which I respond with an input of "Bye" and the logs says 'Adding "Bye" as a response to "Hey"'. But "Bye" was not a response to "Hey" it was a response to "See you later." Any idea why it is doing this? Do I need to add a certain logic adapter or something?

>>> chat("Hey")
INFO:chatterbot.adapters:Received input statement: Hey
INFO:chatterbot.adapters:"Hey" is a known statement
INFO:chatterbot.adapters:Using "Hey" as a close match to "Hey"
INFO:chatterbot.adapters:Selecting response from 8 optimal responses.
INFO:chatterbot.response_selection:Selecting a response from list of 8 options.
INFO:chatterbot.adapters:Response selected. Using "I'm good, thanks!"
INFO:chatterbot.adapters:BestMatch selected "I'm good, thanks!" as a response with a confidence of 1.0
INFO:chatterbot.response_selection:Selecting first response from list of 1 options.
INFO:chatterbot.adapters:LowConfidenceAdapter selected "Huh? I don't get it..." as a response with a confidence of 0
INFO:chatterbot.adapters:NoKnowledgeAdapter selected "Hey" as a response with a confidence of 0
INFO:chatterbot.chatterbot:Adding "Hey" as a response to "Hey"
I'm good, thanks!
>>> chat("Good to hear.")
INFO:chatterbot.adapters:Received input statement: Good to hear.
INFO:chatterbot.adapters:"Good to hear." is not a known statement
INFO:chatterbot.adapters:Using "Good to hear." as a close match to "Goodbye"
INFO:chatterbot.adapters:Selecting response from 1 optimal responses.
INFO:chatterbot.response_selection:Selecting a response from list of 1 options.
INFO:chatterbot.adapters:Response selected. Using "See you later"
INFO:chatterbot.adapters:BestMatch selected "See you later" as a response with a confidence of 0.5
INFO:chatterbot.response_selection:Selecting first response from list of 1 options.
INFO:chatterbot.adapters:LowConfidenceAdapter selected "Huh? I don't get it..." as a response with a confidence of 0
[...]

PR #1606, opened 2019-02-02
Allow in_response_to from input to overried lookup

gunthercox:


iamhssingh:
@vkosuri @gunthercox 
As per pull request #1494 we need to populate `search_in_reponse_to` while learning a response. Probably that is missing here?

gunthercox:
@iamhssingh The `search_in_reponse_to` should be getting populated as long as you are providing it in the input statement. If you have a code snippet or a test case that demonstrates an issue I would be happy to try to correct it.

iamhssingh:
@gunthercox I have commented the issue in #1203 
Here's it: [Link to comment](https://github.com/gunthercox/ChatterBot/issues/1203#issuecomment-475880201)

PR #1612, opened 2019-02-07
Sentence compare alternate impl

capaximperii:
Using word vectors provided by spacy english model "en_core_web_lg" which is a 1 time 1Gb dowload.
Some edits are breaking changes for example the search filters but it is to initiate a discussion on how to create KMeans model 1 time by providing a comparison algorithm access to all data.


gunthercox:
Let me know if you would like me to do a code review at any point. I've been holding off since there appears to be test failures and merge conflicts on this branch. I'd be happy to offer any additional insight that I can if it helps.

capaximperii:
Requires some insight for sure. I will fix the integration issues but first, this is valid only for languages for which spacy provides vectors. So it goes against the chatterbot philosophy of being independent of language. Then, the equivalent sentence can be found very quickly but after the search function returns, the chatterbot engine still takes an (unacceptably) long time to find the response. I have not had much time to see why but looks like something in the sqlite lookup.

gunthercox:
This is an older pull request so I'm going to close it off. Happy to revisit if there is still need for a change here.

PR #1670, opened 2019-03-15
Add pytz to requirements.txt

### LINKED ISSUE #1664: requirements.txt is missing "pytz"
Chatterbot requires **pytz**, but it is not installed with pip.

Running the first example new users currently see this: 

`ModuleNotFoundError: No module named 'pytz'`

florimondmanca:
Fixes #1664. Not sure if `pytz` had to be pinned, if so don't hesitate to point at which version should be used. :+1:


gunthercox:
@florimondmanca This looks good to me. Thank you for the contribution. Hopefully this change helps prevent others from hitting the issue you encountered.

PR #1671, opened 2019-03-19
Requirement to remove warning for SequenceMatcher

fgadaleta:


gunthercox:
Hi @fgadaleta I would prefer to leave this requirement as being optional in ChatterBot. The `python-Levenshtein` library is distributed under the [GPL license](https://github.com/ztane/python-Levenshtein/blob/master/COPYING) which may not be compatible with the way that some developers may choose to use ChatterBot. For example, the GPL license stipulates that a project that uses it must also be licensed under the GPL license and therefore the source code must be made available. Making a project's source code available is not always desirable, such is the case for many commercial applications.

PR #1728, opened 2019-05-17
Remove pymongo from main dependency list

gunthercox:
Pymongo isn't required unless the Mongo DB storage adapter is being used. By not installing pymongo automatically, the installation footprint of ChatterBot is decreased in size, which can be useful when installing on devices that have limited storage space.


vkosuri:
+1

PR #1732, opened 2019-05-20
Fix export_for_training to encode utf8

rafak360:
This makes possible utf8 chat bot's database to be exported.


gunthercox:
@rafak360 Thank you :+1:

Issue #1677, opened 2019-03-24
Two identical strings are not matching


robertcrasmaru:
Hello,

I've tried to add SpecificResponseAdapter to the Django example app and I've encountered the following problem:
The string from statement did not not match with the one from input_text, basically two identical strings did not match.
This line of code generated the problem
https://github.com/gunthercox/ChatterBot/blob/9bd6e57347d2efc1e4b691a147db6fca8bde9851/chatterbot/logic/specific_response.py#L25
Changing it like this fixed it for me
`if statement.text == self.input_text:`

Unfortunately I couldn't understand why this happened since` __str__` method already returns `self.text`. Any idea?

PR #1740, opened 2019-05-30
make can_process function use __str__ of Statement class

minjae1025:
This pull request is intended to solve [this issue](https://github.com/gunthercox/ChatterBot/issues/1677).

It appears that you are trying to compare the member variable `text` and `self.input_text` of statement.

However, this does not call the `__str__` function, it simply compares the id of two objects and will always return false.

Therefore, I call the `str` function to explicitly call the `__str__` function.

I tested this branch using this [test code](https://github.com/gunthercox/ChatterBot/blob/master/examples/specific_response_example.py).

Before, Chatterbot returns invaild answer.
before
After revision, Chatterbot return vaild answer.
after

### DIFF
--- a/chatterbot/logic/specific_response.py
+++ b/chatterbot/logic/specific_response.py
@@ -22,7 +22,7 @@ class SpecificResponseAdapter(LogicAdapter):
         self.response_statement = Statement(text=output_text)
 
     def can_process(self, statement):
-        if statement == self.input_text:
+        if statement.text == self.input_text:
             return True
 
         return False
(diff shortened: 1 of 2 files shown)


chatterbot/logic/specific_response.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, chatbot, **kwargs):
         self.response_statement = Statement(text=output_text)
 
     def can_process(self, statement):
-        if statement == self.input_text:
+        if str(statement) == self.input_text:
```
gunthercox: Since this will always be a statement object, maybe this should be `statement.text` instead of casting to a string?

minjae1025: you are right. Since the str function returns statement.text, there is no problem if statement.text is used. I will fix it

Issue #1730, opened 2019-05-18
Regex issue special character crasher Chatterbot 1.0.5


rafak360:
My settings:
Chatterbot = 1.0.5
Python 3.7.3 (64 bit)
MongoDB 4.0.9 (64 bit)

Everything works fine I can set an entry like 'hello world' whatever, but when using a special character followed by anything such as '(a' or '(z', anything after '(', so I get the exception below, it seems this character '(' has a special meaning in regex, by the way chatterbot version 0.8.7 doesn't have that issue, also when downloading Chatterbot through pip, that misses chatterbot-corpus module which I had to download manually, and that changes PyYAML version to 3.13 from 5.1 (PyYAML version which comes with chatterbot 1.0.5) so I had to get back to 5.1 after getting chatterbot-corpus module.

The following error:
```
[nltk_data] Downloading package stopwords to
[nltk_data]     C:\Users\example\AppData\Roaming\nltk_data...
[nltk_data]   Package stopwords is already up-to-date!
[nltk_data] Downloading package averaged_perceptron_tagger to
[nltk_data]     C:\Users\example\AppData\Roaming\nltk_data...
[nltk_data]   Package averaged_perceptron_tagger is already up-to-
[nltk_data]       date!
Traceback (most recent call last):
  File "D:\MyDocuments\Desktop\Testing.py", line 12, in <module>
    print(chatbot.get_response('hello (a'))
  File "C:\Users\example\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\chatterbot.py", line 139, in get_response
    response = self.generate_response(input_statement, additional_response_selection_parameters)
  File "C:\Users\example\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\chatterbot.py", line 175, in generate_response
    output = adapter.process(input_statement, additional_response_selection_parameters)
  File "C:\Users\example\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\logic\best_match.py", line 29, in process
    closest_match = next(search_results, input_statement)
  File "C:\Users\example\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\search.py", line 74, in search
    for statement in statement_list:
  File "C:\Users\example\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\storage\mongodb.py", line 127, in filter
    kwargs['search_text'] = re.compile(or_regex)
[...]

PR #1745, opened 2019-06-04
Fix regular expression in filter

rafak360:
This Pull Request is addressed to: [this issue](https://github.com/gunthercox/ChatterBot/issues/1730)
When using MongoDB as StorageAdapter, special characters from input messages are triggering regex patterns which is crashing the application.

### DIFF
--- a/chatterbot/storage/mongodb.py
+++ b/chatterbot/storage/mongodb.py
@@ -124,6 +124,7 @@ class MongoDatabaseAdapter(StorageAdapter):
             or_regex = '|'.join([
                 '{}'.format(word) for word in search_text_contains.split(' ')
             ])
+            or_regex = re.escape(or_regex)
             kwargs['search_text'] = re.compile(or_regex)
 
         mongo_ordering = []
--- a/chatterbot/storage/mongodb.py
+++ b/chatterbot/storage/mongodb.py
@@ -122,9 +122,8 @@ class MongoDatabaseAdapter(StorageAdapter):
 
         if search_text_contains:
             or_regex = '|'.join([
-                '{}'.format(word) for word in search_text_contains.split(' ')
+                '({})'.format(word) for word in search_text_contains.split(' ')
             ])
-            or_regex = re.escape(or_regex)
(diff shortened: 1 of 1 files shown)


chatterbot/storage/mongodb.py
```diff
@@ -124,6 +124,7 @@ def filter(self, **kwargs):
             or_regex = '|'.join([
                 '{}'.format(word) for word in search_text_contains.split(' ')
             ])
+            or_regex = re.escape(or_regex)
```
gunthercox: It looks like this change is resulting in a test failure:
```
======================================================================
FAIL: test_search_text_contains_multiple_matches (tests.storage.test_mongo_adapter.MongoAdapterFilterTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/gunthercox/ChatterBot/tests/storage/test_mongo_adapter.py", line 320, in test_search_text_contains_multiple_matches
    self.assertEqual(len(results), 2)
AssertionError: 0 != 2
```

This would need to be corrected before I could merge this.

rafak360: @gunthercox Fixed it, to support regexes with | in them, the expression must be parenthesized. By the way I need your review at: [this pull request](https://github.com/gunthercox/ChatterBot/pull/1732)

rafak360: @gunthercox Dismiss this pull request untill i fix it, my last one didn't work in practice

rafak360: @gunthercox I tried a few commits but didn't work, I don't know how to solve this to be honest, here's something might help: [stackoverflow question](https://stackoverflow.com/questions/280435/escaping-regex-string-in-python)

rafak360: @minjae1025 Maybe you know how to fix it?

rafak360: @gunthercox Fixed, by the way I need your review at: [this pull request](https://github.com/gunthercox/ChatterBot/pull/1732)

PR #1775, opened 2019-07-17
Fix learning_feedback_examply.py

JacoGuerra:
In the older version the feedback was always required, it didn´t matter you replied "yes" or "no", with this version this is fixed and only if you say it´s not a coherent response when you are asked to provide another response.

### DIFF
--- a/examples/learning_feedback_example.py
+++ b/examples/learning_feedback_example.py
@@ -45,7 +45,7 @@ while True:
             response.text,
             input_statement.text
         ))
-        if get_feedback() == False:
+        if get_feedback() is False:
             print('please input the correct one')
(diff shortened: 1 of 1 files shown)


gunthercox:
At a glance, this change looks good to me. But I noticed that at least one test is failing. I can't merge this is until the tests are all passing.

JacoGuerra:
Hi, 

Why one of the test failed? Do I need to do something to pass that test?

Thanks and regards,

gunthercox:
Looks like its just a linting error:
```
./examples/learning_feedback_example.py:48:27:
E712 comparison to False should be 'if cond is False:' or 'if not cond:'
```

PR #1785, opened 2019-07-27
Updated if it gets incorrect response frequently!

shoaibrayeen:
So here-> it was if statment which would check once for incorrect statement. What if it gets incorrect statement twice, more than twice!
Here's the suggested change!
Keep it in while loop. if it gets correct response!
come out of while loop!

The part which needs to be modified!

> while get_feedback():
>             print('please input the correct one')
>             correct_response = Statement(text=input())
>             bot.learn_response(correct_response, input_statement)
>             print('Responses added to bot!')
>             break

### DIFF
--- a/examples/learning_feedback_example.py
+++ b/examples/learning_feedback_example.py
@@ -45,11 +45,12 @@ while True:
             response.text,
             input_statement.text
         ))
-        if get_feedback():
+        while get_feedback():
             print('please input the correct one')
             correct_response = Statement(text=input())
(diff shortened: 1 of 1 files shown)


shoaibrayeen:
> I don't understand the difference between `while` and `if`, the `break` will eventually terminate after first iteration itself.

After an incorrect statement, it should check within learn_response() function -> if it gets the correct output, then it would break otherwise it would take input again and check again.

need to use 'continue' keyword in learn_response() if it gets the incorrect statement.

PR #1839, opened 2019-10-19
Added persona object.

agarwalh673:
Hacktober


gunthercox:
I'm closing this off since this object isn't used anywhere in the code base. No need to create dead code.

PR #1978, opened 2020-05-20
Update tagging.py

qweliant:
Curious if this fix appropriate or if I am implementing incorrectly


gunthercox:
You can also do something like this to link the installed spacy model to the preferred language code name:
```
python -m spacy link en_core_web_sm en
```

Issue #1894, opened 2020-01-08
ModuleNotFoundError: No module named 'chatterbot_corpus'


SaloniSS:
Already tried `pip install --upgrade chatterbot_corpus` and `pip install --upgrade chatterbot`

```
from chatterbot import ChatBot
from chatterbot.trainers import ChatterBotCorpusTrainer

chatbot = ChatBot('Ron Obvious')

# Create a new trainer for the chatbot
trainer = ChatterBotCorpusTrainer(chatbot)

# Train the chatbot based on the english corpus
trainer.train("chatterbot.corpus.english")

# Get a response to an input statement
chatbot.get_response("Hello, how are you today?")
```

sohamnandi007:
Hello @SaloniSS, I would suggest deleting the cache file. 
So, 
Delete the pip folder from **C:\Users\soham\AppData\Local\pip** and
then **pip install chatterbot_corpus**
it will work and remember if it shows **using cached** during installation then it will not install properly in your system so, in those cases remember to delete pip cache files.
**Also, python 3.7.0 + is recommended.**

srianbury:
thanks @sohamnandi007! I didn't realize I needed to `pip install chatterbot_corpus`

Ahammedjaleel:
Try below command 
pip install git+git://github.com/gunthercox/chatterbot-corpus@master#egg=chatterbot_corpus

Issue #1757, opened 2019-06-15
No module named 'chatterbot_corpus' when migrate


hustjiangtao:
cmd: 
```bash
# install
pip install chatterbot

# django settings
INSTALLED_APPS = (
    # ...
    'chatterbot.ext.django_chatterbot',
)

# run migrate
python manage.py migrate django_chatterbot
```

python version: `3.6.5`

```
Traceback (most recent call last):
  File "manage.py", line 21, in <module>
    main()
  File "manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 357, in execute
    django.setup()
  File "/usr/local/lib/python3.6/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.6/site-packages/django/apps/registry.py", line 91, in populate
    app_config = AppConfig.create(entry)
  File "/usr/local/lib/python3.6/site-packages/django/apps/config.py", line 116, in create
    mod = import_module(mod_path)
  File "/usr/local/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 941, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/local/lib/python3.6/site-packages/chatterbot/__init__.py", line 4, in <module>
    from .chatterbot import ChatBot
  File "/usr/local/lib/python3.6/site-packages/chatterbot/chatterbot.py", line 2, in <module>
[...]

Issue #1727, opened 2019-05-15
An error after installation


boumediane10:
I'm getting this error after the installation

C:\Users\Programmer\Desktop\ChatterBot-master\examples>python basic_example.py
Traceback (most recent call last):
  File "basic_example.py", line 1, in <module>
    from chatterbot import ChatBot
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\__init__.py", line 4, in <module>
    from .chatterbot import ChatBot
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\chatterbot.py", line 2, in <module>
    from chatterbot.storage import StorageAdapter
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\storage\__init__.py", line 1, in <module>
    from chatterbot.storage.storage_adapter import StorageAdapter
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\storage\storage_adapter.py", line 3, in <module>
    from chatterbot.tagging import PosHypernymTagger
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\tagging.py", line 4, in <module>
    from chatterbot.tokenizers import get_sentence_tokenizer
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\tokenizers.py", line 4, in <module>
    from chatterbot.corpus import load_corpus, list_corpus_files
  File "C:\Users\Programmer\AppData\Local\Programs\Python\Python37\lib\site-packages\chatterbot\corpus.py", line 5, in <module>
    from chatterbot_corpus.corpus import DATA_DIRECTORY
ModuleNotFoundError: No module named 'chatterbot_corpus'

Issue #1708, opened 2019-04-22
ModuleNotFoundError: No module named 'chatterbot_corpus' error


SperCoder:
i was running a the basic example code from \examples:

`

**_from chatterbot import ChatBot
from chatterbot.trainers import ListTrainer

# Create a new chat bot named Charlie
chatbot = ChatBot('Charlie')

trainer = ListTrainer(chatbot)

trainer.train([
    "Hi, can I help you?",
    "Sure, I'd like to book a flight to Iceland.",
    "Your flight has been booked."
])

# Get a response to the input text 'I would like to book a flight.'
response = chatbot.get_response('I would like to book a flight.')

print(response)

`_**

________
Your work is great so pls respond fastly! Thanks!

PrinceChukz:
I am having the same issue of no module chatterbot corpus, must i pip install chatterbot-corpus to be able to train the chatterbot? Or will pip install chatterbot alone do everything

SperCoder:
I fixed it myself. pip install chatterbot-coprus worked for me

Orfeous:
Hello @PrinceChukz,
@SperCoder is correct.
The corpus itself is a separate package, thus have to be installed separately.

PrinceChukz:
pip install chatterbot-corpus worked but it uninstalls the version of the pyYAML file needed as a requirement for the chatterbot to work properly and installs an older version. Which is not good for the training

PrinceChukz:
![image](https://user-images.githubusercontent.com/37223482/56588003-6eefdc80-65da-11e9-8951-fb2bac49cc7f.png)

PrinceChukz:
@gunthercox @Orfeous Getting that error when i try installing from git as stated in chatterbot requirements.txt

Orfeous:
Hello @PrinceChukz,

This looks strange. 
Just to close out common cases:

- Do you have enough free disk space?
- What version of windows 10 ou running? The path may be longer that 260 characters. [Read more here](https://www.howtogeek.com/266621/how-to-make-windows-10-accept-file-paths-over-260-characters/)
- when you browse into that directory can you write and/or modify files there?

shamsher786:
Not able to install chatterbot-corpus library using pip command.

![Screenshot_71](https://user-images.githubusercontent.com/20358682/59011386-33d7fe80-8852-11e9-9e41-0f6f24e221f9.png)

[...]

PR #2005, opened 2020-07-07
Patch ModuleNotFoundError (No module named 'chatterbot_corpus')

TEEN-BOOM:
While this can be patched by installing chatterbot_corpus I feel we should just define the variable here, and prevent people in the future from opening issues regarding the same.

Issues this patches: #1894 , #1852, #1757, #1727, #1689,  #1708, #833

Also chatterbot_corpus is not installed by pip, for some reason

### DIFF
--- a/chatterbot/corpus.py
+++ b/chatterbot/corpus.py
@@ -2,7 +2,10 @@ import os
 import io
 import glob
 import yaml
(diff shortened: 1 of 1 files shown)


gunthercox:
@TEEN-BOOM this looks like a great solution. Thank you for contributing this.

PR #2026, opened 2020-08-20
Fixing "Conversation 2" label

luantorrex:
The second block was repeating the label "Conversation 1",
As you can see, I fixed it as "Conversation 2" and converted into a SVG file as the previous one.

training-graph


gunthercox (changes requested):
It looks like this change might have modified the vector graphic from being a collection of vector elements into being an embedded base64 encoded image. I kind of prefer the elements since they seem cleaner and more easily editable.

What was this pull request suppose to fix, it isn't clear in the provided description?

luantorrex:
You are right @gunthercox it seem cleaner and more easily editable, but look at the image from your repo https://github.com/gunthercox/ChatterBot/blob/master/docs/_static/training-graph.svg and the image from my pull request, at the second block you repeated the label "Conversation 1", I fixed this label.

gunthercox:
Sorry, I never got around to re-reviewing this pull request but I recently found the original source file for the diagram:

image

I included the update in #2403

PR #2078, opened 2020-12-03
Relax Python version requirement

fridex:
Fixes installation when Python version has a patch release:

```
ERROR: Package 'ChatterBot' requires a different Python: 3.8.3 not in '>=3.4, <=3.8'
```


xloem (approved):

gunthercox:
I'm working on some changes starting with PR #2389 that should eventually lead to wider Python version support. I've started by dropping support for older versions, up next: adding support for newer ones.

PR #2094, opened 2021-01-03
Update languages.py

jzsmoreno:
OSError: [E050] Can't find model 'en'. It doesn't seem to be a shortcut link, a Python package or a valid path to a data directory. SOLVED

### DIFF
--- a/chatterbot/languages.py
+++ b/chatterbot/languages.py
@@ -581,7 +581,7 @@ class ELX:
 
 
 class ENG:
-    ISO_639_1 = 'en'
+    ISO_639_1 = 'en_core_web_sm'
     ISO_639 = 'eng'
     ENGLISH_NAME = 'English'
 


chatterbot/languages.py
```diff
@@ -581,7 +581,7 @@ class ELX:
 
 
 class ENG:
-    ISO_639_1 = 'en'
+    ISO_639_1 = 'en_core_web_sm'
```
gunthercox, line 584: `languages.py` is a collection of classes representing language codes, this line should continue to be the ISO_639 language code. I've made a change in #2389 that I believe will resolve this with a bit more flexibility for different languages:
https://github.com/gunthercox/ChatterBot/pull/2389/files#diff-a30358b68e779ddc228075ea1421f86901a2f3818447a4d905624706b43f42c7R30

PR #2109, opened 2021-02-01
Django 3

dancaron:
The example Django app needed a few updates to be compatible with Django 3. Specifically: {% load staticfiles %} was deprecated in Django 2.1, and removed in Django 3.0.


arihant2math:
@dancaron Can you fix the failing tests?

PR #2121, opened 2021-02-15
New preprocessor - remove punctuation

Belobragin:
New preprocessor to remove punctuation.

It is based on this answer:
https://stackoverflow.com/questions/15547409/how-to-get-rid-of-punctuation-using-nltk-tokenizer

### DIFF
--- a/chatterbot/preprocessors.py
+++ b/chatterbot/preprocessors.py
@@ -2,6 +2,21 @@
 Statement pre-processors.
 """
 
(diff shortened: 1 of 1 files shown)


gunthercox:
My apologies for never getting around to reviewing this when it was origionally opened. It's been a while, but I think there might be limited use cases for this preprocessor so I'm going to close this off. Happy to reconsider if any consistent requests for it emerge.

PR #2147, opened 2021-04-20
Avoid to load a model from a shortcut, which is deprecated as of spacy 3.0

lucia-w:


gunthercox:
I've made a different change as a part of #2389 that should achieve a similar effect with a bit more flexibility for other languages: https://github.com/gunthercox/ChatterBot/pull/2389/files#diff-5fad1f9c3be573497a8f25e97994e2475e240f7034eb59a04d6cec8137efbbd9R156-R163

PR #2175, opened 2021-06-30
Python_requires 3.8 syntax 

bobloy:
`<=3.8` explicitly only allows 3.8.0, and disallows 3.8.1 and above.

This fixes it by swapping it with `<3.9`


arihant2math:
I would say python 3.9 is okay too.

bobloy:
Turns out this is a duplicate of #2078

PR #2191, opened 2021-08-19
Update requirements.txt

kritarth2121:
It is proper Requirements.txt ,preivous one is giving errors


gunthercox:
That is a lot of new requirements 😄

If it helps, I've made changes as a part of #2389 to upgrade this projects requirements.

PR #2195, opened 2021-08-28
checking the version pf pip

anukaal:


gunthercox:
This was a good idea to add, but newer versions of `pip` appear to have this built in now and will prevent installation when a package requires specific Python versions.

Example:

```bash
INFO: pip is looking at multiple versions of chatterbot to determine which version is
compatible with other requirements. This could take a while.
ERROR: Package 'chatterbot' requires a different Python: 3.10.16 not in '~=3.11'
```

PR #2206, opened 2021-10-10
fixed bug SpaCy OSError: Can't find model 'en'

Rohitw3code:
approve this , this cause a lot of trouble in deployment of project made with chatterbot


Diego448 (commented):
Although it seems to fix the immediate issue, this only works based on the assumption that only English language will be used and all other languages will still fail, and it is not taking into account other parts of the library that also needs to load the nlp (natural language processing) modules as in comparisons.py module

gunthercox:
Hey, thanks for the pull request. I've made some different updates as a part of #2389 that should have a similar effect, so I'm going to close this one off.

PR #2220, opened 2021-11-24
Fixed errors related with spacy models not found

Diego448:
For the most common issues related with the error "osError: [E050] Can't find model 'en'" I created this pull request which has an updated version on the languages module with a new field called `SPACY_MODEL` which contains the new Spacy model names and will be used for both `tagging` and `comparisons` modules instead of the `ISO_639_1` field only if available.

*Important notes regarding this pull request:*
  - SPACY_MODEL field of language module only contains the small variant of the related language not the large and more accurate one, which is smaller, faster and less precise.
  - Only supported languages (those which already have models for them in the official Spacy page) have the SPACY_MODEL field, not supported ones will have no such field

### DIFF
--- a/chatterbot/comparisons.py
+++ b/chatterbot/comparisons.py
@@ -74,7 +74,10 @@ class SpacySimilarity(Comparator):
             )
             raise OptionalDependencyImportError(message)
 
-        self.nlp = spacy.load(self.language.ISO_639_1)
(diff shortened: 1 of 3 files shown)


gunthercox:
Hey, thanks for the pull request. I've made some different updates as a part of #2389 that should have a similar effect, so I'm going to close this one off.

Issue #1983, opened 2020-05-25
django.core.exceptions.ImproperlyConfigured: 'chatterbot.ext.django_chatterbot' must supply a name attribute.


YuzhouZhang:
After I integrate with Django, I got this exception when I execute "python manager.py runserver":

> Traceback (most recent call last):
>   File "C:\Users\mamajingjie\AppData\Local\Programs\Python\Python36\lib\site-packages\django\apps\config.py", line 149, in create
>     app_name = cls.name
> AttributeError: type object 'AppConfig' has no attribute 'name'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "manage.py", line 21, in <module>
>     main()
>   File "manage.py", line 17, in main
>     execute_from_command_line(sys.argv)
>   File "C:\Users\mamajingjie\AppData\Local\Programs\Python\Python36\lib\site-packages\django\core\management\__init__.py", line 381, in execute_from_command_line
>     utility.execute()
>   File "C:\Users\mamajingjie\AppData\Local\Programs\Python\Python36\lib\site-packages\django\core\management\__init__.py", line 357, in execute
>     django.setup()
>   File "C:\Users\mamajingjie\AppData\Local\Programs\Python\Python36\lib\site-packages\django\__init__.py", line 24, in setup
>     apps.populate(settings.INSTALLED_APPS)
>   File "C:\Users\mamajingjie\AppData\Local\Programs\Python\Python36\lib\site-packages\django\apps\registry.py", line 91, in populate
>     app_config = AppConfig.create(entry)
>   File "C:\Users\mamajingjie\AppData\Local\Programs\Python\Python36\lib\site-packages\django\apps\config.py", line 152, in create
>     "'%s' must supply a name attribute." % entry)
> django.core.exceptions.ImproperlyConfigured: 'chatterbot.ext.django_chatterbot' must supply a name attribute.

SystemDZ:
You confused class name with it's attribute

in your **settings.py** then **INSTALLED_APPS**
change **AppConfig** to **ApiConfig**

toffeeandlisa:
I don't know if answering this after one year is against the rules of github, but you are an anonimous hero. Thanks.

PR #2230, opened 2022-01-08
fix bug in tagging.py with spacy en model

Omarzintan:
This PR fixes a bug in the `chatterbot/tagging.py` file

# Implementation
* A description of the bug and solution can be found in this stack overflow [thread](https://stackoverflow.com/questions/66087475/chatterbot-error-oserror-e941-cant-find-model-en)
  * Basically one of the lines in tagging.py is updated to cater for when chatterbot tries to load a model titled `en`. It ensures that the correct naming of the English model is used, i.e. `en_core_web_sm`

### DIFF
--- a/chatterbot/tagging.py
+++ b/chatterbot/tagging.py
@@ -23,7 +23,10 @@ class PosLemmaTagger(object):
 
         self.punctuation_table = str.maketrans(dict.fromkeys(string.punctuation))
 
-        self.nlp = spacy.load(self.language.ISO_639_1.lower())
+        if self.language.ISO_639_1.lower() == 'en':
+            self.nlp = spacy.load('en_core_web_sm')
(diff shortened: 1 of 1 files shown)


NoMoreDreams (approved):
This probably fixes only the en model

NoMoreDreams (approved):

pretty-girl123 (commented):
#1983

pretty-girl123:
hi everyone how are you doing?

gunthercox:
I've made a similar change in #2389 that I believe will resolve this with a bit more flexibility for different languages:
https://github.com/gunthercox/ChatterBot/pull/2389/files#diff-a30358b68e779ddc228075ea1421f86901a2f3818447a4d905624706b43f42c7R30

Issue #1208, opened 2018-02-13
AttributeError: 'tuple' object has no attribute 'confidence'


ravaga:
Im trying to set up a custom logic adapter, but somehow im getting
`AttributeError: 'int' object has no attribute 'confidence'`
im following the exact example provided on the docs.. could someone point out what im doing wrong..

This is my return of the def process function on my logic adapter

        if response.status_code == 200:
            confidence = 1
        else:
            confidence = 0

        info = data.get('data', 'unavailable')

        response_statement = Statement('The current info is {}'.format(info))
        return confidence, response_statement

ravaga:
Nevermind.. i figured out... 
```
response_statement = Statement('The current info is {}'.format(info))
response_statement.confidence = confidence
return response_statement
```

vkosuri:
@ravaga glad to figured it out :)

PR #2267, opened 2022-09-05
Fix error in return variable documentation

martin-martin:
If `.process()` in your custom Logic Adapter returns a tuple with `(confidence, return_statement)`, like currently described in the docs, then running the bot gives an error:

```
AttributeError: 'tuple' object has no attribute 'text' chatterbot
```

This was addressed also in https://github.com/gunthercox/ChatterBot/issues/1208 and it's probably worth to update the docs to reflect that `.confidence` is an attribute on a `Statement()`.


vignesh1507 (approved):

gunthercox (approved):
Thank you for updating this

PR #2309, opened 2023-05-09
tagging.py changes for spacy>=3.0 versions

Comder101:


gunthercox:
Hey, thanks for the pull request. I've made some different updates as a part of #2389 that will have a similar effect, so I'm going to close this one off.

PR #2336, opened 2023-11-26
updated the Spacy dependency version

krishnan0005:
updated the Spacy dependency version which was blocking installing using PIP


gunthercox:
Hey, thanks for the pull request. I've made some different updates as a part of #2389 that will have a similar effect, so I'm going to close this one off.

PR #2366, opened 2024-04-24
Fixed the version issue and added the Chinese version README

WThirteen:
Fix the issue that the spacy version is incompatible；
Added Chinese version of README；
Added a test file;


README.md
```diff
@@ -1,5 +1,10 @@
 ![ChatterBot: Machine learning in Python](https://i.imgur.com/b3SCmGT.png)
-
```
WThirteen, line 2: Added tags for Chinese and English reademe

chatterbot/tagging.py
```diff
@@ -23,7 +23,12 @@ def __init__(self, language=None):
 
         self.punctuation_table = str.maketrans(dict.fromkeys(string.punctuation))
 
-        self.nlp = spacy.load(self.language.ISO_639_1.lower())
```
WThirteen, line 26: Fix the problem that occurs in the spacy version

WThirteen (commented):
Any problems that may arise are resolved

PR #2376, opened 2024-09-20
Updated the python files by removing the typos and added more comments to understand the code more easily.

vignesh1507:
I've made some changes in the codes as it had few typos in it. Added few comments in the code too.


vignesh1507:
Updated setup.py file with the newer verison of python i.e 3.12.

PR #2388, opened 2025-01-04
Configure GitHub Actions CI workflow

gunthercox:


gunthercox:
Alright, tests and linting are now mostly set back up. Since Travis CI appears to be paid-only now GitHub Actions seems like the way to go.

PR #2413, opened 2025-05-20
Switch from char field to text field types

gunthercox:


gunthercox:
Not all databases support `TextField`s with indexes applied to them, I'll hold off on making this change.

PR #2444, opened 2026-06-07
Add repeating character preprocessor

ShreyasB1:
Add a preprocessor that reduces runs of three or more repeated letters
down to two (e.g. "sooooo" -> "soo"). Elongated words are common in
conversational input, and normalizing them maps spelling variations to a
consistent form so the chat bot can match input against trained
statements more reliably.

Repeated digits and punctuation are left unchanged, and naturally
occurring double letters (such as the "oo" in "cool") are preserved.

Includes unit tests and documentation for the new preprocessor.

### DIFF
--- a/chatterbot/preprocessors.py
+++ b/chatterbot/preprocessors.py
@@ -3,7 +3,7 @@ Statement pre-processors.
 """
 from chatterbot.conversation import Statement
 from unicodedata import normalize
-from re import sub as re_sub
+from re import sub as re_sub, compile as re_compile
 from html import unescape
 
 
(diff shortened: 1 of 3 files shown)


jocelyn1981:
邮件已收到，谢谢您！

ShreyasB1:
Hi! I was wondering the timeline for this feature being approved.

chatterbot/preprocessors.py
```diff
@@ -45,3 +45,27 @@ def convert_to_ascii(statement: Statement) -> Statement:
...
+    Elongated words are common in conversational text (for example
+    "I am sooooo happy"). Collapsing the repeated characters maps these
+    variations to a single, consistent form ("I am soo happy") which helps
+    the chat bot match input against statements it has been trained on.
+
+    Letter pairs that occur naturally (such as the "oo" in "cool") are
+    preserved, and repeated digits or punctuation are left unchanged.
+    """
+    statement.text = _REPEATING_CHARACTER_PATTERN.sub(
+        lambda match: match.group(1) * 2, statement.text
```
gunthercox, line 68: Taking a look at some related sample data, it appears that this regex would group a statement such as `"I am sooooo happy"` with `"I am soo happy"` (x2 characters), which still doesn't quite reach the intended token of `"so"`.

I'm not certain this works as intended in some of the cases the pull request was expecting. Perhaps there is another approach that might work better? (If not, a project-specific preprocessor is always an alternative option to including one in the main codebase.

syedkosaainhaider-maker:
ok

chatterbot/preprocessors.py
ShreyasB1: Good catch — you're right, and I reworked it.

I went looking for a way to tell `"sooooo"` → `"so"` apart from `"gooood"` → `"good"`, and the honest answer is that it needs a dictionary. I checked whether spaCy's vocab could serve as one, and it can't: with `en_core_web_sm`, `"soo"` and `"col"` both report as present in `nlp.vocab.strings`, and `is_oov` is `True` for every token because the small model ships no vectors. Adding a real word list felt out of scope for a preprocessor.

So instead of reducing the run to two characters, it's now reduced to one:

- `"I am sooooo happy"` → `"I am so happy"`
- `"Yesss that was greaaaat"` → `"Yes that was great"`

This is safe because the pattern only matches runs of three or more, and no correctly spelled English word has one. Checking against `/usr/share/dict/words` (235,976 entries), only 7 match, all archaic `-ssship` constructions (`bossship`, `goddessship`, `wallless`). So the preprocessor never alters text that was already spelled correctly. Naturally doubled letters are runs of two, so `"cool"` is still untouched — the two-character version was protecting a case that was never at risk.

The tradeoff that remains: a word that genuinely contains a doubled letter is reduced past its correct spelling when elongated, so `"gooood"` → `"god"`. That's documented in the docstring and pinned by a test, with a note that a project needing the distinction can register its own preprocessor backed by a word list.

Also fixed the flake8 errors (E305/E302/W293) that were in my earlier commits.

Happy to close this and keep it project-specific instead if you'd still rather not carry the ambiguity in core.

jocelyn1981:
邮件已收到，谢谢您！

PR #2449, opened 2026-08-22
Fix symlink containment in UbuntuCorpusTrainer.extract()

gunthercox:


arpitjain099:
Ran the race against the updated branch. It holds.

Same harness as before, swapping `data_directory` for a symlink once extraction has visibly started, 30 runs at each size:

```
members=   50  attacker won 0/30 runs, 0 files written outside
members=  500  attacker won 0/30 runs, 0 files written outside
members= 4000  attacker won 0/30 runs, 0 files written outside
```

Against 24/30, 27/30 and 27/30 before, with 1168, 13472 and 107971 files landing outside. The swap itself still landed on all 90 runs, so that is the anchoring doing the work rather than the harness missing its window.

I also forced the flag off to exercise the non-dir_fd path, since that is what macOS and the BSDs take: 0/30 at every size there as well. So the staging directory plus the atomic rename closes it on its own, and the fd anchoring is defence in depth rather than the load-bearing part. Worth knowing if you ever need to simplify.

One thing to fix before this merges, which I opened as #2451: `_DIR_FD_EXTRACTION_SUPPORTED` is assigned twice and the second assignment drops the `/proc/self/fd` term, so the flag comes out true on macOS where procfs does not exist. The `mkdir` succeeds because it is genuinely fd-relative, then tarfile writes through `/proc/self/fd/N` and `extract()` dies on the first write with `OSError: [Errno 30] Read-only file system: '/proc'`.

